PR #4, opened 2025-08-07
Add interoperability with Workers built-in RPC

kentonv:
This turns out to be easier than I expected. We don't need a mechanism for registering types, we can just make JSRPC's stubs extend RpcTarget.

There are some bugs in workerd, though, fixed by this PR: https://github.com/cloudflare/workerd/pull/4719

So the tests will fail for now. Once we get a new workerd release out with those changes then I can update the package here and the tests will pass.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /node_modules
 notes.txt
+/dist/


kentonv:
Overrode workerd to 2025-08-09 which has the bug fixes, tests pass, yay.

PR #18, opened 2025-09-21
Improve docs, and some other tweaks

kentonv:


kentonv:
Ironically the code de-dupe commit _added_ three bytes to the gzipped size. :sob:

PR #22, opened 2025-09-22
Fix types in React example

third774:
The React example had a few type errors — I _think_ this is the right way to fix? Awesome package! 🎉


examples/worker-react/web/src/main/App.tsx
```diff
@@ -1,5 +1,6 @@
-import React, { useMemo, useState } from 'react'
+import { useMemo, useState } from 'react'
```
third774, line 1: Just removing React import as it shows as unused (and isn't necessary as of React >17)

kentonv:
I guess @southpolesteve doesn't receive github notifications, oh well.

kentonv:
Thanks!

PR #27, opened 2025-09-23
failing test: can an answer promise roundtrip

kumavis:
encountered this edgecase while working on the OCapN js implementation, was curious how capnweb handled it.

There's room for argument on what the correct behavior is, but I find the current error message indicates a bug more than an appropriate assertion.

```
 FAIL   browsers-without-using (webkit)  __tests__/index.test.ts > roundtrip pipeline value handling > can return a pipelined promise unchanged
Error: Attempted to use RPC stub after it has been disposed.
 ❯ module code src/core.ts:241:14
    239| 
    240| const DISPOSED_HOOK: StubHook = new ErrorStubHook(
    241|     new Error("Attempted to use RPC stub after it has been disposed."));
       |              ^
    242| 
    243| // A call interceptor can be used to intercept all RPC stub invocations wit…

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[6/6]⎯
```

### DIFF
--- a/__tests__/test-util.ts
+++ b/__tests__/test-util.ts
@@ -62,4 +62,9 @@ export class TestTarget extends RpcTarget {
   returnNull() { return null; }
   returnUndefined() { return undefined; }
   returnNumber(i: number) { return i; }
+
+  echo(arg: any) {
+    return arg;
+  }
+
 }
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -1403,3 +1403,21 @@ describe("MessagePorts", () => {
         new Error("Peer closed MessagePort connection."));
   });
 });
+
+describe("Roundtripped answer promise handling", () => {
+  it("can use a roundtripped answer promise", async () => {
+    const stub = new RpcStub(new TestTarget());
+
+    // Create a pipelined promise
+    let counterPromise = stub.makeCounter(5);
+    
+    // Echo the pipelined promise back
+    let echoedPromise = stub.echo(counterPromise);
+    
+    // Both should resolve to the same counter
+    let [counter, echoedCounter] = await Promise.all([counterPromise, echoedPromise]);
+
+    expect(await counter.increment()).toBe(6);
+    expect(await echoedCounter.increment()).toBe(7);
+  });
+});


kumavis:
seems you can't echo any reference; this is not unique to answers. is this a bug?

__tests__/test-util.ts
```diff
@@ -62,4 +62,9 @@ export class TestTarget extends RpcTarget {
   returnNull() { return null; }
   returnUndefined() { return undefined; }
   returnNumber(i: number) { return i; }
+
+  echo(arg: any) {
+    return arg;
```
kentonv, line 67: So, there's an ownership problem here:
* Any stubs in an RPC's params are owned by the caller. They will be auto-disposed upon return.
* Returning a stub from an RPC transfers ownership to the caller.

[The rules are documented here.](https://github.com/cloudflare/capnweb?tab=readme-ov-file#automatic-disposal)

But this means that when you `return arg` here, the caller now has double-ownership.

What ends up happening is, the parameter is disposed upon return, but that means that the returned value is now invalid.

You can solve this by:

```suggestion
    return arg.dup();
```

I'll admit I'm not thrilled with this state of affairs. But I don't know what else we can do. It's important that args are auto-disposed upon return, because otherwise stubs that the callee wasn't expecting to receive will always be leaked. It's also important that returns transfer ownership, because otherwise it's difficult to dispose a stub _after_ returning it.

Maybe we need to expose a helper that walks an object tree and dups all the stubs, so that this function can be written in a way that doesn't assume anything about the type of `arg`.

kentonv:
Closing this since it's working as intended.

PR #43, opened 2025-09-24
show and tell: rough sketch of third party handoff

kumavis:
rough draft of third-party handoff for capnweb based on my experience writing an OCapN js implementation

### current status
basic test passes. no further work planned atm

### Basic Architecture
- Allows for opt-in Third Party Handoff support by having each client create a `SessionManager` and pass it to each session via `RpcSessionOptions`
- When a `SessionManager` has been specified and an import from third session is to be sent (eg via call args or return value), it instead:
  - generates a "swissnum" (shared secret) used for registering and redeeming a gift
  - send a "registerGift" message to the reference's host: `["registerGift", exportId, swissnum]`
  - serializes the reference as a "gift" type that species its host and swissnum: `["gift", locator, swissnum]`
  - upon receipt of a "gift" type, sending "redeemGift" message to the gift host: `["redeemGift", swissnum]`
  - upon receipt of a "redeemGift" message, resolving to registered gift value

### Implementation Details
 - each `SessionManager` is constructed with a `connectHook` for establishing a session from a locator
 - Adds a `getLocator()` method to `RpcTransport` to present a location string for the protocol + destination.
 - required making the target extractable from TargetStubHook. I couldn't find any other way of getting a PayloadStubHook to resolve to a TargetStubHook without throwing.

### Notes
- does not introduce 3-way e-order (neither does OCapN)
- likely requires all participating peers to have a SessionManager and be able to connect to a given locator
- OCapN handoffs sign the handoff for the recipient, here I'm assuming swissnum security is sufficient

### Not implemented
- handling `redeemGift` arriving before `registerGift` (due to networking latency)
- handoff of promises
- did not pay attention to hook cleanup / disposal
- session cleanup in SessionManager

### DIFF
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -7,6 +7,7 @@ import { deserialize, serialize, RpcSession, type RpcSessionOptions, RpcTranspor
          RpcStub, newWebSocketRpcSession, newMessagePortRpcSession,
          newHttpBatchRpcSession} from "../src/index.js"
 import { Counter, TestTarget } from "./test-util.js";
+import { SessionManager } from "../src/core.js";
 
 let SERIALIZE_TEST_CASES: Record<string, unknown> = {
   '123': 123,
@@ -112,6 +113,10 @@ class TestTransport implements RpcTransport {
   private aborter?: (err: any) => void;
   public log = false;
 
+  getLocator(): string {
+    return `test://${this.name}`;
+  }
+
   async send(message: string): Promise<void> {
     // HACK: If the string "$remove$" appears in the message, remove it. This is used in some
     //   tests to hack the RPC protocol.
@@ -157,11 +162,12 @@ class TestHarness<T extends RpcTarget> {
 
   stub: RpcStub<T>;
 
-  constructor(target: T, serverOptions?: RpcSessionOptions) {
-    this.clientTransport = new TestTransport("client");
-    this.serverTransport = new TestTransport("server", this.clientTransport);
+  constructor(target: T, serverOptions?: RpcSessionOptions, clientOptions?: RpcSessionOptions, label?: string) {
+    const prefix = label ? `${label}:` : "";
+    this.clientTransport = new TestTransport(`${prefix}client`);
+    this.serverTransport = new TestTransport(`${prefix}server`, this.clientTransport);
 
-    this.client = new RpcSession<T>(this.clientTransport);
+    this.client = new RpcSession<T>(this.clientTransport, undefined, clientOptions);
 
     // TODO: If I remove `<undefined>` here, I get a TypeScript error about the instantiation being
     //   excessively deep and possibly infinite. Why? `<undefined>` is supposed to be the default.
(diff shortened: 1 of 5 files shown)


kentonv:
Thanks for thinking about this.

But, we need to think carefully about the protocol design before jumping to an implementation.

Some important considerations:

----------------

First, Cap'n Proto also implements 3PH, and we'd like to bridge Cap'n Web to Cap'n Proto, so we'll want them to agree on how it works.

----------------

Second, we should consider whether swissnums are the best answer here, or if we should allow custom transports to use other authentication mechanisms, such as public/private key pairs. Cap'n Proto's implementation actually allows different transports to plug in different authentication mechanisms. That said, it might be overengineered.

----------------

Third, sending the locator with the "gift" is not always ideal. Sometimes you want the capability's provider to be the one that initiates the connection to the recipient.

In particular, this is commonly true in tail-calling scenarios:

```ts
// alice
await bob.foo();

// bob
foo() { return carol.foo(); }

// carol
foo() { return 123; }
```

In this scenario:
1. Alice calls Bob
2. Bob calls Carol, and tells Carol to register the result as a gift.
3. Concurrently with 2, Bob tells Alice to accept a gift from Carol.
4. Alice send a message to Carol to accept the gift.
5. Carol returns the result.

Let's assume we have a magical network where establishing a connection is zero-RTT. Even then, in this interaction, the total time taken to get the result is the greater of:
a. One round trip between Alice and Bob (1&3), PLUS one round trip between Alice and Carol (4&5).
b. One three-way round trip Alice -> Bob -> Carol -> Alice.

In almost all scenarios, (a) will be longer, and so that will determine the latency. But _in theory_, it feels like (b) is what we really want.

On the other hand, imagine we re-arrange the protocol such that instead of Alice sending "redeemGift" to Carol, we have Carol send "provideGift" to Alice. This also requires that "registerGift" contains a locator, so that Carol can connect to Alice.

Then we have:
1. Alice calls Bob
2. Bob calls Carol, and tells Carol to send the result to Alice.
3. Carol sends the result to Alice.
4. Concurrently with 2&3, Bob tells Alice to expect a gift from Carol.

The total time here is the maximum of:
a. One round trip between Alice and Bob.
b. One three-way round trip Alice -> Bob -> Carol -> Alice.

In this case, we probably expect (b) is always greater that (a). But (b) is what we want. We win!

But the standard gifting model is still best when sending a capability, as opposed to making a call. That is:
* For "exports", the standard gifting model is best.
* For "questions", the reverse gifting model is best.

----------------

Fourth, there is an issue with gift chains. Imagine:

```ts
// alice
await bob.foo();

// bob
foo() { return carol.foo(); }

// carol
foo() { return dave.foo(); }

// dave
foo() { return 123; }
```

If you use the standard gifting model, then:
1. Alice calls Bob.
2. Bob tells Alice to accept a gift from Carol (while simultaneously registering the gift with Carol).
3. Alice calls Carol to accept the gift.
4. Carol tells Alice to accept a gift from Dave.
5. Alice calls Dave to accept the gift.
6. Dave returns the result.

Now our total time is the sum of round trips Alice<->Bob, Alice<->Carol, Alice<->Dave. Yuck! We want a single loop Alice->Bob->Carol->Dave->Alice.

If we use the "reverse" gifting model discussed above, then we have:
1. Alice calls Bob.
2. Bob tells Alice to expect a gift from Carol (while simultaneously registering the gift with Carol).
3. When Carol forwards to Dave, Carol also sends a message to Alice saying "Here's your gift: Please expect a gift from Dave".
4. Finally, Dave provides the actual result gift.

In theory, the overall latency here is the single loop Alice->Bob->Carol->Dave->Alice. However, the message in step 3 seems wasteful, especially in real-world conditions: Carol forms a whole connection to Alice, just to tell Alice that the gift is coming from Dave instead.

We can optimize this: Bob can tell Alice: "Expect that _someone_ will give you a gift with this swissnum." And then Carol never contacts Alice at all, but rather instructs Dave to send the gift to Alice.

A similar problem presents itself in the forward gifting case: If a capability is sent from Alice to Bob to Carol to Dave, it would be nice if Carol does not have to connect to Alice simply to "regift" the capability.

This stuff actually matters for Cloudflare Workers, in particular for Durable Object routing. (Cap'n Proto's implementation of 3PH actually covers this already.)

kumavis:
thanks for the thorough response/
I initiated this mostly as an educational exercise in learning capnweb rather than an earnest PR. Agree that it deserves careful design.

Funny you mention pushing handoff auth into the Transport/NetLayer, just had a similar conversation with some Ocapn folks.
I'm mulling a layered standard of a pairwise protocol and a 3PHO protocol, but to layer the implementations properly you need a membrane but can use the serialization boundary as that membrane if theres a way to hook in there.

I'll chew on "reverse gifting" for a bit

zarutian:
See provideFor and acceptFrom methods of [NonceLocator](http://wiki.erights.org/wiki/NonceLocator) of CapTP of E for some inspiration.

kumavis:
A couple notes from a conversation with markm (http://erights.org) on reverse gifting:
- if you dont specify the recipient of the gift, the GC window of the gift entry is undefined.
- relying purely on swissnums requires confidentiality, making it not suitable for some environments like blockchains

kumavis:
closing this show-and-tell PR as it has served its purpose 👍

kentonv:
> if you dont specify the recipient of the gift, the GC window of the gift entry is undefined.

You still have a vine that you drop once the handoff is complete.

kumavis:
> You still have a vine that you drop once the handoff is complete.

i think what markm was trying to convey is that if no one responds to resolve your question, you have to just let that question sit around uncollected (no connection/session to tie the life cycle to)

PR #57, opened 2025-09-25
docs: add deno server example to the README

sigmaSd:
if I uncomment headers it gives this type error 

```
TS2353 [ERROR]: Object literal may only specify known properties, and 'headers' does not exist in type 'RpcSessionOptions'.
        headers: { "Access-Control-Allow-Origin": "*" }, // TODO: this is a type error
        ~~~~~~~
    at file:///home/mrcool/dev/deno/lab/zz/a.ts:26:9
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -595,7 +595,7 @@ class MyApiImpl extends RpcTarget implements MyApi {
   // ... define API, same as above ...
 }
 
-Deno.serve((req) => {
+Deno.serve(async (req) => {
   const url = new URL(req.url);
   if (url.pathname === "/api") {
     if (req.headers.get("upgrade") === "websocket") {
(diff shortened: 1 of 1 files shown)


sigmaSd:
Maybe the request should be updated instead if this function really doesn't accept headers 

```ts
      // If you are accepting WebSockets, then you might as well accept cross-origin HTTP, since
      // WebSockets always permit cross-origin request anyway. But, see security considerations
      // for further discussion.
      const headers = new Headers(req.headers);
      headers.set("Access-Control-Allow-Origin", "*");
      const updatedReq = new Request(req, { headers });
      return newHttpBatchRpcResponse(updatedReq, new MyApiImpl());
```

sigmaSd:
I updated the code as you suggested

kentonv:
Thank you!

sigmaSd:
Thank you as well!

PR #62, opened 2025-09-27
 fix: Add warning for RPC target instance property access (fixes #55)

### LINKED ISSUE #55: `RpcTarget` properties resolved in runtime is not what the type suggests
```typescript
import { RpcTarget, RpcStub, newMessagePortRpcSession } from "capnweb";

interface MyApi extends RpcTarget {
  myProp: number;
}

class MyApiImpl extends RpcTarget implements MyApi {
  myProp: number = 1;
}

const { port1, port2 } = new MessageChannel();

using stub: RpcStub<MyApi> = newMessagePortRpcSession<MyApi>(port1);

using _sessionFrom2To1 = newMessagePortRpcSession(port2, new MyApiImpl());

const myProp: number = await stub.myProp;

console.log(myProp);
// Expect: 1
//    Got: undefined

const myPropStr: string = await stub.myProp.toString();
console.log(myPropStr);
// Expect: "1"
//    Got: [object Function]
```

You can test this by running `deno test.ts` or `node test.ts` with `Node >= 24`

Bethel-nz:
This PR fixes Issue #55.

### Problem

When a developer attempts to access an instance (or "own") property on an `RpcTarget`, the library correctly returns `undefined` as per its security policy. However, this behavior can be confusing as there is no feedback explaining why it's not working.

### Solution

This change introduces a `console.warn` message when an instance property is accessed over RPC.

This improves the developer experience by:
- Clearly indicating why the property access returns `undefined`.
- Maintaining the existing behavior of returning `undefined`.
- Suggesting the use of getters as the recommended approach, as discussed in the issue.

### Testing

A new test file (`test/issue55.test.ts`) has been added to verify that the new warning is logged correctly and that the property still resolves to `undefined`.

### DIFF
--- a/__tests__/issue55.test.ts
+++ b/__tests__/issue55.test.ts
@@ -1,0 +1,111 @@
+// Copyright (c) 2025 Cloudflare, Inc.
+// Licensed under the MIT license found in the LICENSE.txt file or at:
+//     https://opensource.org/license/mit
+
+import { expect, it, describe, beforeEach, afterEach } from "vitest";
+import { RpcTarget, RpcStub, newMessagePortRpcSession } from "../src/index.js";
+
+describe("Issue #55: RPC target property access", () => {
+  let originalWarn: typeof console.warn;
+  let warnings: string[];
+
+  beforeEach(() => {
+    // Spy on console.warn to capture warnings
+    warnings = [];
+    originalWarn = console.warn;
+    console.warn = (message: string) => {
+      warnings.push(message);
+    };
+  });
+
+  afterEach(() => {
+    // Restore original console.warn
+    console.warn = originalWarn;
+  });
+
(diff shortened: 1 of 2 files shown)


kentonv:
It seems you ran some sort of auto-formatter on the code as part of this PR. Please don't do that.

More broadly, we'd generally prefer that people not submit PRs unless we specifically ask. For an issue like this, it actually takes more work to review an external contribution than it takes for the team to make the fix directly, especially when there isn't agreement upfront about the right solution.

Here, I don't think logging a warning on the server is the right thing to do. It is the client that is doing something wrong, so it is the client that needs to be informed. Probably we should just throw an exception in this case, rather than returning undefined.

I'm going to close this and let's return discussion to the issue thread.

Issue #53, opened 2025-09-25
"Type instantiation is excessively deep and possibly infinite" when calling a method on an untyped session


tom-sherman:
To reproduce, construct a session (either ws or http) without passing a stub into the generic arg.

For http, simply constructing the session is enough to get the error.

```ts
import { newHttpBatchRpcSession } from "capnweb";
const http = newHttpBatchRpcSession("...");
```

For ws, the error occurs when attempting to call a method:

```ts
import { newWebSocketRpcSession } from "capnweb";
const api = newWebSocketRpcSession("...")
api.ping();
```

Expected behaviour:

I'd expect for it to return something like `Stub<unknown>` and then error with when accessing any methods. Essentially I should get an untyped stub that I can pass around.

kentonv:
Thanks, I've noticed this too but haven't had a chance to fully debug yet. The types are a little bit buggy still unfortunately.

vicary:
This makes `deno check` throws `Maximum call stack size limit exceeded` at random files, depending on how many files are check at once.

Spent like $50 on Opus 4.5 to dig it up, cause it's breaking CI pipelines.

PizzaConsole:
also having problems with types in a very large project
`Type instantiation is excessively deep and possibly infinite`

PR #63, opened 2025-09-27
fix(types): Add default generic for untyped RPC sessions

Bethel-nz:
**Resolves Issue #53**

This PR fixes a TypeScript compilation error that occurred when creating RPC sessions without an explicit type parameter.

## Problem

When `newHttpBatchRpcSession()` was called without a generic type, the TypeScript compiler would fail with the error:

```
"Type instantiation is excessively deep and possibly infinite.ts(2589)"
```

This happened because the function's generic `T extends Serializable<T>` did not have a default type. When TypeScript tried to infer `T`, the recursive nature of the `Serializable<T>` constraint caused an infinite loop in the type checker.

## Reproduction

```typescript
// Before fix - TypeScript compilation fails
const http = newHttpBatchRpcSession("..."); // ❌ Type instantiation error
```

## Solution

This PR aligns the behavior of `newHttpBatchRpcSession` with `newWebSocketRpcSession` by adding `= Empty` as the default for the generic type parameter `T`.

```typescript
// src/index.ts
- export let newHttpBatchRpcSession<T extends Serializable<T>>
+ export let newHttpBatchRpcSession<T extends Serializable<T> = Empty>
```

This simple change provides a concrete type for TypeScript to use when one isn't specified, breaking the recursive loop and resolving the compilation error.

## Result

After this fix, both HTTP and WebSocket sessions can be created without explicit types, providing a consistent developer experience.

```typescript
// Both now work as expected for untyped sessions
const httpSession = newHttpBatchRpcSession("http://example.com"); // ✅ Works
const wsSession = newWebSocketRpcSession("ws://example.com");     // ✅ Still works

// Explicitly typed sessions are unaffected and continue to work correctly
interface MyApi { 
  ping(): Promise<string>; 
}
const typedHttp = newHttpBatchRpcSession<MyApi>("http://example.com"); // ✅ Still works
```

## Testing

Comprehensive test coverage has been added to `test/issue53.test.ts` to:

- ✅ Verify that both untyped and typed sessions can be created without errors
- ✅ Confirm that untyped sessions return the expected `RpcStub<Empty>` behavior
- ✅ Ensure that explicitly typed sessions continue to work as expected, preventing any regressions

### DIFF
--- a/__tests__/issue53.test.ts
+++ b/__tests__/issue53.test.ts
@@ -1,0 +1,66 @@
+// Copyright (c) 2025 Cloudflare, Inc.
+// Licensed under the MIT license found in the LICENSE.txt file or at:
+//     https://opensource.org/license/mit
+
+import { expect, it, describe } from "vitest";
+import { newHttpBatchRpcSession, newWebSocketRpcSession } from "../src/index.js";
+
+describe("Issue #53: Type instantiation excessively deep for untyped sessions", () => {
+  it("should not cause type instantiation errors for untyped HTTP batch sessions", () => {
+    // Before the fix, this would cause:
+    // "Type instantiation is excessively deep and possibly infinite.ts(2589)"
+    const httpSession = newHttpBatchRpcSession("http://example.com");
+    expect(httpSession).toBeDefined();
+
+    // RPC stubs are proxied functions, so they appear as "function" type
+    expect(typeof httpSession).toBe("function");
+  });
+
+  it("should not cause type instantiation errors for untyped WebSocket sessions", () => {
+    // This should work (already had default type parameter)
+    const wsSession = newWebSocketRpcSession("ws://example.com");
+    expect(wsSession).toBeDefined();
+
+    // RPC stubs are proxied functions, so they appear as "function" type
+    expect(typeof wsSession).toBe("function");
+  });
+
+  it("should work with explicit type parameters (typed sessions)", () => {
+    // Explicit typing should continue to work as before
+    interface MyApi {
+      ping(): Promise<string>;
+      getData(): Promise<number>;
+    }
+
+    const typedHttp = newHttpBatchRpcSession<MyApi>("http://example.com");
+    const typedWs = newWebSocketRpcSession<MyApi>("ws://example.com");
+
+    expect(typedHttp).toBeDefined();
+    expect(typedWs).toBeDefined();
+    expect(typeof typedHttp).toBe("function");
+    expect(typeof typedWs).toBe("function");
+
+    // These should have the correct types (checked at compile time)
+    // The calls themselves would fail at runtime since we're not connected to real servers,
+    // but TypeScript should accept them syntactically
+    expect(typeof typedHttp.ping).toBe("function");
+    expect(typeof typedWs.getData).toBe("function");
+  });
+
+  it("should return Empty-typed stubs for untyped sessions", () => {
+    // The untyped sessions should return RpcStub<Empty> which behaves like unknown
+    const httpSession = newHttpBatchRpcSession("http://example.com");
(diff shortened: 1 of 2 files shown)


kentonv:
This function is not intended to be called without a type. If the type is unknown, you should use `<unknown>`, or perhaps `<any>`.

`newWebSocketRpcSession()` permits omitting the type because it is a symmetrical function -- it can be used on either the client or the server. When used on the server, typically you expect that the main interface exported by the client is indeed empty, because normally it's only the server that exports a main interface to the client.

`newHttpBatchRpcSession()`, however, can only be used by the client, and is useless if the server exports an empty interface, since the client will have nothing to call in the batch.

Bethel-nz:
Thanks for explaining the design philosophy - that makes perfect sense! The distinction between symmetrical vs client-only functions is a great insight.

I tried both suggested alternatives, but they hit compilation issues:
* `<unknown>`: `Type 'unknown' does not satisfy the constraint 'Serializable<unknown>'`
* `<any>`: Same infinite recursion error

This puts developers in a tough spot - they want to follow your guidance but can't due to the constraint design.

But would you prefer an overload approach instead? Something like:

```typescript
// Helpful error when no type provided
export function newHttpBatchRpcSession(...): {
  Error: 'newHttpBatchRpcSession requires a type argument. e.g. newHttpBatchRpcSession<MyApi>(...)'
};

// Properly typed version  
export function newHttpBatchRpcSession<T extends Serializable<T>>(...): RpcStub<T>;

// Implementation
export function newHttpBatchRpcSession(
  urlOrRequest: string | Request,
  init?: RequestInit & RpcSessionOptions
): any {
  return newHttpBatchRpcSessionImpl(urlOrRequest, init);
}
```

This approach:
- Maintains the design philosophy (requires explicit types)
- Provides clear guidance instead of confusing infinite recursion errors
- Uses standard TypeScript patterns
- Minimizes the infinite recursion error

It doesn't solve the underlying recursive constraint issue, but it does limit the problematic usage patterns for clients and guides them toward the correct API usage.

I believe the recursion stems from the `Serializable<T>` type definition being inherently self-referential - when TypeScript tries to check `T extends Serializable<T>`, it gets caught in a loop evaluating the nested `Serializable<T[K]>` checks. An alternative approach might be to separate the type-level constraint from the runtime serialization logic, perhaps using structural typing or marker interfaces instead of the recursive constraint pattern.

**Side note**: I also noticed a potential issue where the client-side `newHttpBatchRpcSession` accepts `RequestInit` options (including fetch options like headers), but the server-side `newHttpBatchRpcResponse` only receives the final `Request` object and `RpcSessionOptions`. This means custom headers and other fetch options get lost in transmission, even though the types suggest they should work. Not sure if this is intentional or if there's a missing piece in the request forwarding logic.

kentonv:
I suspect there is a more fundamental change we need to make to the types to fix these infinite recursion errors. This isn't the only place where it comes up.

The types were based on types written for Workers RPC originally, and although I made a number of changes for Cap'n Web, I have to admit I still don't fully grok them.

> Side note: I also noticed a potential issue where the client-side newHttpBatchRpcSession accepts RequestInit options (including fetch options like headers), but the server-side newHttpBatchRpcResponse only receives the final Request object and RpcSessionOptions. This means custom headers and other fetch options get lost in transmission, even though the types suggest they should work. Not sure if this is intentional or if there's a missing piece in the request forwarding logic.

I don't see the problem. `newHttpBatchRpcSession()` lets you specify request options so you can add headers or other options for your app's use in routing the request. Cap'n Web itself does not care about these headers.

kentonv:
#142 fixed #53 and many other issues, and has landed in Cap'n Web 0.6.0. I'm going to assume this PR is no longer needed and close it, but feel free to open a new issue if there's still a problem.

Issue #61, opened 2025-09-27
Bun Support Missing


BurkeDesigns:
How do I get capnweb to work in bun? I recieve the client message, but no server side function is executed.

<img width="1641" height="1348" alt="Image" src="https://github.com/user-attachments/assets/4817152b-9abb-4df2-bc52-050e6a8b7343" />

BurkeDesigns:
After much experimentation (trial and error) I discovered the following:

The issue is that Bun's WebSocket object doesn't support addEventListener in the same way as a standard WebSocket, so I wrapped it to store the event listeners and manually trigger them in the message handler. As this is very messy and hacky, could you guy's fix that in the capnweb library itself to handle that automatically? 

The screenshot below is how I finally got it to work in Bun.

<img width="1641" height="1348" alt="Image" src="https://github.com/user-attachments/assets/1316432d-7b8c-4890-85d1-4b700f957cba" />

kentonv:
Oof, it seems like Bun has chosen a deeply non-standard and weirdly procedural-style design for their WebSocket server-side API. I guess supporting this would require implementing a custom RpcTransport just for Bun.

Or I suppose you could use Bun's `node:http` compat, but I imagine that's not a very palatable solution?

BurkeDesigns:
@kentonv Yeah, that would be less than ideal to use "node:http".

balazskosi:
Here is the skeleton how I use capnweb with bun:

```typescript
const webServer = Bun.serve<BunWebSocketTransport, any>({
    port,
    routes: {
        "/connect": req => {                    
            if (webServer.upgrade(req, {data: new BunWebSocketTransport()})) {
                return;
            }
            return new Response("Upgrade failed", {status: 500});
        },                
    },
    websocket: {
        open: ws => ws.data.onOpen(ws),
        message: (ws, message: string | Buffer) => ws.data.onMessage(message),
        close: (ws, code, reason) => ws.data.onClose(code, reason)
    }
});
```

```typescript
import type {RpcTransport} from "capnweb";
import type {ServerWebSocket} from "bun";

export default class BunWebSocketTransport implements RpcTransport {
    #websocket: ServerWebSocket<any> | undefined;
    #sendQueue : string[] | undefined = [];
    #receiveResolver?: (message: string) => void;
    #receiveRejecter?: (err: any) => void;
    #receiveQueue: string[] = [];
    #error?: any;

    onOpen(websocket: ServerWebSocket<any>): void {
[...]

PR #65, opened 2025-09-29
chore: add Hono examples

yusukebe:
This PR adds new examples. They will show how to implement Cap'n Web servers on Cloudflare Workers, Deno, and Node.js with Hono.


examples/hono/src/deno.ts
```diff
@@ -0,0 +1,23 @@
...
+import { MyApiServer } from './my-api-server.ts'
+
+const app = new Hono()
+
+app.get(
+  '/api',
+  upgradeWebSocket((_c) => {
+    return {
+      onOpen(_event, ws) {
+        newWebSocketRpcSession(ws.raw!, new MyApiServer())
```
kentonv, line 13: I imagine this doesn't work on Bun, since it doesn't use the standard `WebSocket` API under the hood. Is that right?

I wonder if it makes sense to write a custom transport specifically for Hono, that uses Hono's wrapper and so works with all runtimes Hono supports?

Although I see Hono's `upgradeWebSocket` is actually different for each runtime, so perhaps this code can't really be totally runtime-independent anyway?

What do you think?

threepointone: there are plenty of hono-* packages, so hono-capnweb can totally be a thing (or even officially as a `@hono/` package

yusukebe: Ah, there is a ready #61! [The approch](https://github.com/cloudflare/capnweb/issues/61#issuecomment-3404005663) by @grampelberg is great. With it, we can make Cap'n Web support Bun independent of Hono. Let's discuss it on #61.

We, Hono side, will develop "Cap'n Web Middleware" as `@hono/capnweb` in our repo ([honojs/middleware](https://github.com/honojs/middleware)), not in this repo.

Regarding this PR, I'll revise after creating `@hono/capnweb`.

yusukebe:
Hi. We released `@hono/capnweb`:

https://github.com/honojs/middleware/tree/main/packages/capnweb

If you want to use Cap'n Web with Hono, you can use it. So, these examples are no longer needed. I'll close this PR now, but we may have to update the README to point out `@hono/capnweb`. Anyway, thanks.

PR #74, opened 2025-10-07
Fixes for protocol definition

### LINKED ISSUE #42: callArguments in protocol.md does not actually appear to be an expression
Per [protocol.md](https://github.com/cloudflare/capnweb/blob/49c3543e185231a98fef7782977006bcece46dbd/protocol.md), describing the "import"/"pipeline" expression types:

> `callArguments` is also optional. If specified, then the given property should be called as a function.
> **`callArguments` is an expression that evaluates to an array**; these are the arguments to the call.

(emphasis added)

However, it appears this isn't the case, for example, the following code:

```javascript
const api = newHttpBatchRpcSession(RPC_URL);
let result = await api.hello("date", 12345);
```

Sends the following:

```
["push",["pipeline",0,["hello"],["date",12345]]]
["pull",1]
```

Of note, if we look at the 4th field in the pipeline expression, we see `["date",12345]`. If we attempt to interpret that value as an expression, it will evaluate to a single date object, not an array of arguments.

Instead, I believe `callArguments` is actually an array of expressions, not an expression that evaluates to an array.

### LINKED ISSUE #68: Recommendation: Clarify array escaping rules in Cap’n Web spec
### Issue

The current specification for “literal arrays” leaves room for multiple interpretations. It is not defined whether arrays nested inside objects (or deeper structures) must be escaped, or if escaping applies only to top-level arrays. This ambiguity can lead to inconsistent behavior across implementations.

### Recommendation

Update the specification to state that all arrays, regardless of context, must be escaped by wrapping them in an outer array.

Examples:

- [1,2,3] → [[1,2,3]]
- [] → [[]]
- {"items": [1,2]} → {"items": [[1,2]]}
- {"data": {"items": []}} → {"data": {"items": [[]]}}
 

This approach provides a single, unambiguous rule, aligns with the TypeScript reference implementation, and ensures consistent behavior across languages.

### Context

While developing the Rust implementation (capn-rs), my initial serializer generated unescaped arrays inside objects. To test this approach, I patched the TypeScript reference implementation to accept them. Although workable, this added complexity and diverged from the reference behavior. After exploring both paths, I found that escaping arrays at all levels is simpler, consistent, and a better fit for interoperability.

Dr-Emann:
Fix the protocol description of array escaping to clarify its recursive nature, and fix the docs around the type of callArguments in a import or pipeline expression.

Fixes #42 
Fixes #68
Closes #69


kentonv:
Sorry for the long delay on reviewing this -- my last month was jam-packed with other stuff and I failed to get to Cap'n Web. Finally getting a chance now.

This looks like a good doc update, thank you!

PR #75, opened 2025-10-08
Spelling

jsoref:


__tests__/index.test.ts
```diff
@@ -270,20 +270,20 @@ describe("local stub", () => {
     expect(await outerStub.innerTarget.square(3)).toBe(9);
   });
 
-  it("returns undefined when accessing non-existent properties", async () => {
+  it("returns undefined when accessing nonexistent properties", async () => {
```
jsoref, line 273: cloudflare is [somewhat inconsistent](https://github.com/search?q=org%3Acloudflare+nonexistent&type=code) , but it's worth noting that at least [_some_ of the time it uses this spelling](https://github.com/cloudflare/cloudflare-docs/blob/63541abea1fabd8345c79e4d1a95aa2aa87ccad8/src/content/docs/dns/dns-firewall/random-prefix-attacks/about.mdx?plain=1#L19-L21)

kentonv:
Thanks.

PR #78, opened 2025-10-08
Export the `Serializable` type from the package.

itaylor:
This is needed to aide integration with other libraries. For instance, if you want to make a generic function that will make use of an RpcStub, you will need to declare a constraint of `Serializable<T>`, but you currently can't because Serializable isn't exported.

The woraround is to copy the entire type definition and declare it locally, which works, but this is much simpler.

### DIFF
--- a/.changeset/dirty-bikes-glow.md
+++ b/.changeset/dirty-bikes-glow.md
@@ -1,0 +1,5 @@
+---
+"capnweb": minor
+---
+
+The package now exports the type `RpcCompatible<T>` (previously called `Serializable<T>`, but not exported), which is needed when writing generic functions on `RpcStub` / `RpcPromise`.


kentonv:
Makes sense but I wonder if we want to change the name to `RpcCompatible` before making this public?

After all pass-by-reference types aren't really serialized.

kentonv:
(Sorry for taking so long to look at this -- had a busy month.)

itaylor:
I agree, `Serializable` isn't a good name for what this does.  `RpcCompatible` is a better name.

itaylor:
I've updated the PR to change the name to `RpcCompatible`

kentonv:
Thanks, I'd love to merge this but it looks like there's now a conflict, can you resolve?

itaylor:

itaylor:
I've rebased this on top of latest main branch

PR #82, opened 2025-10-11
Try a different approach to importing "cloudflare:workers".

### LINKED ISSUE #23: Many bundlers don't like the conditional import of "cloudflare:workers" and the top-level await
the `workersModule` import in core.ts uses a top-level await. Causing the following error:
```
Module format "iife" does not support top-level await. Use the "es" or "system" output formats rather.
```

I'm trying this out for a browser extension I'm working on using a custom RpcTransport on top of browser.runtime.onConnect. However the framework I'm using [wxt](https://wxt.dev) uses iife for building the content script files, es isn't natively supported.

kentonv:
Fixes #23

Per #23, many bundlers do not like our dynamic import of "cloudflare:workers". This PR takes a different approach, using the "exports" feature of "package.json" to point workerd at a different version of the code. That version imports "cloudflare:workers" and sticks it in the global scope before importing the rest of the library, thus allowing it to conditionally probe.

@threepointone is this a sane approach? I have very little understanding of JS tooling but this seem like it works.

### DIFF
--- a/__tests__/test-server-workerd.js
+++ b/__tests__/test-server-workerd.js
@@ -10,7 +10,7 @@
 // build step for it. Instead, we're getting by configuring the worker in vitest.config.ts by
 // just specifying the raw JS modules.
 
-import { newWorkersRpcResponse } from "../dist/index.js";
+import { newWorkersRpcResponse } from "../dist/index-workers.js";
 import { RpcTarget, DurableObject } from "cloudflare:workers";
 
 // TODO(cleanup): At present we clone the implementation of Counter and TestTarget because
--- a/__tests__/workerd.test.ts
+++ b/__tests__/workerd.test.ts
@@ -5,7 +5,7 @@
 /// <reference types="@cloudflare/workers-types" />
 import { expect, it, describe } from "vitest";
 import { RpcStub as NativeRpcStub, RpcTarget as NativeRpcTarget, env, DurableObject } from "cloudflare:workers";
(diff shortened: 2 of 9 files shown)


threepointone:
Missed this, will see over the weekend

scotthellewell:
This pull request fixes all the issues I was having.  Any idea when it will be merged and a new package created?

threepointone:
looking at it today

kentonv:
@scotthellewell Sorry for the delay on this and everything else Cap'n Web! In the month since Cap'n Web was released a whole lot of stuff has conspired to take of my time (and Sunil's too), and this week I'm on vacation. Next week I plan to circle back and address all the open Cap'n Web issues.

scotthellewell:
@kentonv I hope you have a great vacation.  I've grabbed your branch and compiled my own package from it for now which works.  I just like to directly use from the npm repository when possible.  But great library so far.

scotthellewell:
Awesome to see this is approved now.  Do you have an estimate on when this will get merged and published to npm?  I'd love to just reference your package, but I need this fix.  If it will be a while still, I can publish a separate scoped npm package to use for the time being. But if it will be done in the next few days I would prefer to just wait.  I would love an ETA.  Thanks.

kentonv:
Yes I'm planning to do a pass through all the open issues tomorrow and then probably push a release.

Meant to do it yesterday but there was some other fire that came up. :/

Issue #93, opened 2025-10-29
`ReadableStreams` and `Response` objects are not supported


dario-piotrowicz:
Passing `ReadableStreams` (and as a result `Response` objects as well) is currently not supported

This is a known limitation as [described in the README](https://github.com/cloudflare/capnweb/blob/c3409357b1c84dd515b4e739786addbdd135c244/README.md?plain=1#L209) however I am creating this issue to hopefully track this and see if there is any progress on potentially supporting them.

Also I am opening this off of https://github.com/cloudflare/workers-sdk/issues/11114 (since that is not a wrangler limitation but a capnweb one).

pliashkou:
Hey @kentonv , thanks for providing more information in the PR. I agree with the response. Would you mind, please, sharing your vision of how it should work when you have a bit of time? Thanks again!

benfoxall:
I'm interested in `ReadableStream` support. 

But I was also wondering if this functionality might work as an async generator/iterator?

```js
class MyApiServer extends RpcTarget {
  hello(name) {
    return `Hello, ${name}!`
  }
  async * countUp() {
     let i = 0;
     while(true) {
        await new Promise(r => setTimeout(r, 1000))
        yield i++;
     }
  }
}


// on the client
for await(const i of api.countUp()) {
   console.log(`counter: ${i}`)

   if(i > 10) break;
}
```

nakasyou:
I found how to respond AsyncIterator, so I share it.
```ts
// server
export class MyRpc extends RpcTarget {
  stream() {
    const stream = new ReadableStream({
      start(controller) {
        controller.enqueue('h')
        controller.enqueue('e')
        controller.enqueue('l')
        controller.enqueue('l')
        controller.enqueue('o')
        controller.enqueue('!')
        controller.close()
      },
    })[Symbol.asyncIterator]()
    return {
      next: stream.next.bind(stream),
      return: stream.return?.bind(stream),
      throw: stream.throw?.bind(stream),
    }
  }
}
```
```ts
// client
const sess = newWebSocketRpcSession<MyRpc>('..')

const res = await sess.stream()

for await (const chunk of {
  [Symbol.asyncIterator]() {
    return res
  }
}) {
  console.log('Received chunk:', chunk)
}
```

Result:

<img width="198" height="143" alt="Image" src="https://github.com/user-attachments/assets/879f0bee-320a-4514-bc8a-b54533797d08" />

ericclemmons:
[...]

PR #94, opened 2025-10-29
feature: streams implementation

pliashkou:
This PR solves [this Issue](https://github.com/cloudflare/capnweb/issues/93) and adds support of Readable and Writable streams to capnweb.

### DIFF
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -1413,7 +1413,6 @@ describe("ReadableStream and WritableStream support", () => {
     using clientStub = newMessagePortRpcSession<TestTarget>(channel.port2);
 
     // Create a readable stream on the server and receive it on the client
-    // The stream is automatically converted - no helper function needed!
     let stream = await clientStub.createReadableStream([1, 2, 3, 4, 5]);
 
     // Read all chunks from the stream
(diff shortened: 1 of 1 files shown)


kentonv:
Hi @pliashkou,

Thanks for trying to help -- but this isn't the right way to support streams.

In order to be efficient over a network connection, it's critical that streams implement flow control. In this implementation, you have converted read() and write() into RPCs that call to the other side. The problem with this is that it means every read and write call will require a network round trip.

Imagine a connection with 100ms of latency (pretty common), and imagine that you are reading/writing 4k chunks (also pretty common). Then, this implementation would only be able to achieve 40kps total bandwidth -- extremely low. We need to arrange for multiple reads / writes to run concurrently. But it's also important not to create _too many_ concurrent requests as this can clog up the socket buffer, creating excessive latency for other RPCs on the same connection.

I have a plan to support streams property, but the details are subtle. It's best if I do it. I'm hoping to implement this in the next couple weeks. (Sorry I've been quiet over the last month -- I've had a ton of other stuff going on, including a fire drill around some silly benchmarks, the Cloudflare Connect conference, some vacation, etc. -- but I think I can finally devote some time to Cap'n Web again now.)

Sorry that this means this work has gone to waste. For future contributions, I recommend starting by opening a discussion where we can make sure we're on the same page on design details, before writing code.

PR #99, opened 2025-11-01
Add support for additional serialization types

lmaccherone:
# Add Support for Additional Serialization Types

Adds serialization support for additional JavaScript types to align with Cloudflare Workers RPC capabilities.

### Special Numbers
`["special-number", "NaN"|"Infinity"|"-Infinity"]`

### JavaScript Built-in Types
- **RegExp**: `["regexp", {source: string, flags: string}]`
- **Map**: `["map", [[[key1, val1]], [[key2, val2]], ...]]`
- **Set**: `["set", [[val1], [val2], ...]]`
- **ArrayBuffer**: `["arraybuffer", "base64string"]` (same encoding as `Uint8Array`/`"bytes"`, different tag to preserve type)

### Web API Objects
- **URL**: `["url", url.href]`
- **Headers**: `["headers", [[key, val], ...]]`

### Enhanced Error Support
New format: `["error", {name: string, message: string, stack?: string, cause?: unknown, customProps?: Record<string, unknown>}]`

Backward compatible with old format: `["error", name, message, stack?]`

Error serialization follows existing `onSendError` hook behavior (stack included only if hook returns an Error with stack property). New format adds support for `cause` chain and custom properties.

## Implementation

All changes follow existing Cap'n Web patterns. Type detection uses prototype matching (or `instanceof` for `URL`/`Headers`). Serialization/deserialization follows the existing `["type", data]` array format. Recursive types (Map, Set) use the same 64-level depth limit as objects and arrays.

## Format Examples

```javascript
// Map - entries wrapped in arrays (following Cap'n Web's array-wrapping convention)
new Map([["foo", "bar"]]) → ["map", [[["foo","bar"]]]]

// Set - array values wrapped
new Set(["foo", "bar"]) → ["set", ["foo","bar"]]

// ArrayBuffer - base64 with padding removed
new ArrayBuffer(4) → ["arraybuffer", "AAAA"]

// Headers - normalized to lowercase
headers.set("Content-Type", "application/json") → ["headers", [["content-type","application/json"]]]
```

## WebKit Test Failures

Two pre-existing test failures exist in the WebKit browser environment (HTTP requests and WebSocket connections). These are unrelated to serialization changes and appear to be Playwright/WebKit compatibility issues with localhost connectivity. Tests pass in Chromium, Firefox, Node.js, and workerd environments. We are running these tests on a MacBook Pro M4 running macOS 15.6.1 (24G90).

## Notes

- **Backward Compatible**: Old serialized data (e.g., old Error format) deserializes correctly. New serializations automatically use upgraded formats.
- **No Dependencies**: Code is inlined, maintaining Cap'n Web's "<10kB with no dependencies" goal.
- **No Cycle Support**: Explicitly does not support cycles/aliases (as desired by Cap'n Web authors).
- **Browser-Specific Error Properties**: Filtered out for consistent cross-browser behavior (WebKit: `line`, `column`, `sourceURL`; Firefox: `fileName`, `lineNumber`, `columnNumber`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,20 +194,22 @@ let userId: number = await authedApi.getUserId();
 
 The following types can be passed over RPC (in arguments or return values), and will be passed "by value", meaning the content is serialized, producing a copy at the receiving end:
 
-* Primitive values: strings, numbers, booleans, null, undefined
+* Primitive values: strings, numbers (including `NaN`, `Infinity`, and `-Infinity`), booleans, null, undefined
 * Plain objects (e.g., from object literals)
 * Arrays
 * `bigint`
 * `Date`
 * `Uint8Array`
-* `Error` and its well-known subclasses
+* `ArrayBuffer`
+* `Map` and `Set`
+* `RegExp`
+* `URL` and `Headers`
+* `Error` and its well-known subclasses (with full-fidelity serialization including `cause` chains and custom properties)
 
 The following types are not supported as of this writing, but may be added in the future:
-* `Map` and `Set`
-* `ArrayBuffer` and typed arrays other than `Uint8Array`
-* `RegExp`
+* Typed arrays other than `Uint8Array`
 * `ReadableStream` and `WritableStream`, with automatic flow control.
-* `Headers`, `Request`, and `Response`
+* `Request` and `Response` (require asynchronous body handling)
 
 The following are intentionally NOT supported:
 * Application-defined classes that do not extend `RpcTarget`.
@@ -317,7 +319,7 @@ To facilitate interoperability:
 So basically, it "just works".
 
 With that said, as of this writing, the feature set is not exactly the same between the two. We aim to fix this over time, by adding missing features to both sides until they match. In particular, as of this writing:
-* Workers RPC supports some types that Cap'n Web does not yet, like `Map`, streams, etc.
+* Workers RPC supports some types that Cap'n Web does not yet, like streams.
 * Workers RPC supports sending values that contain aliases and cycles. This can actually cause problems, so we actually plan to *remove* this feature from Workers RPC (with a compatibility flag, of course).
 * Workers RPC does not yet support placing an `RpcPromise` into the parameters of a request, to be replaced by its resolution.
 * Workers RPC does not yet support the magic `.map()` method.
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -26,9 +26,25 @@ let SERIALIZE_TEST_CASES: Record<string, unknown> = {
   '["date",1234]': new Date(1234),
   '["bytes","aGVsbG8h"]': new TextEncoder().encode("hello!"),
   '["undefined"]': undefined,
-  '["error","Error","the message"]': new Error("the message"),
-  '["error","TypeError","the message"]': new TypeError("the message"),
-  '["error","RangeError","the message"]': new RangeError("the message"),
+  '["error",{"name":"Error","message":"the message"}]': (() => { let e = new Error("the message"); delete e.stack; return e; })(),
+  '["error",{"name":"TypeError","message":"the message"}]': (() => { let e = new TypeError("the message"); delete e.stack; return e; })(),
+  '["error",{"name":"RangeError","message":"the message"}]': (() => { let e = new RangeError("the message"); delete e.stack; return e; })(),
+  '["special-number","NaN"]': NaN,
+  '["special-number","Infinity"]': Infinity,
+  '["special-number","-Infinity"]': -Infinity,
+  '["regexp",{"source":"test","flags":"gi"}]': /test/gi,
+  '["regexp",{"source":"^\\\\d+$","flags":""}]': /^\d+$/,
+  '["map",[[["foo","bar"]]]]': new Map([["foo", "bar"]]),
+  '["map",[[["a","b"]],[["c","d"]]]]': new Map([["a", "b"], ["c", "d"]]),
+  '["set",["foo","bar"]]': new Set(["foo", "bar"]),
+  '["arraybuffer","aGVsbG8h"]': new TextEncoder().encode("hello!").buffer,
+  '["url","https://example.com/path?q=1"]': new URL("https://example.com/path?q=1"),
+  '["headers",[["content-type","application/json"],["x-custom","value"]]]': (() => {
+    let h = new Headers();
+    h.set("Content-Type", "application/json");
+    h.set("X-Custom", "value");
+    return h;
+  })(),
 };
 
 class NotSerializable {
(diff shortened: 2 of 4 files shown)


kentonv:
Hi Larry,

Thanks for working on this.

This PR description sounds AI-generated. Is it? I don't have a problem with AI code, but I do prefer that AI-generated content be clearly marked as such, since it calls for a different approach to reviewing the code.

As for the content, the reason many of these types weren't supported already is because there are nuances to the question of whether we want to support them at all, and if so, how exactly they should work. I think we need to answer those before getting into implementation.

* inf/nan: This was an oversight and I actually added support in 0.2.0 which was just released. The encoding is just `["inf"]`, `["-inf"]`, and `["nan"]` -- I initially considered adding something like the "special-number" grouping but I couldn't convince myself that it provided any actual benefit.
* RegExp: I'm not really convinced that anyone actually cares to pass this type over RPC, and I am concerned because it is a fairly complex type with a weird interface that could lead to surprises if it were sent in a context where it wasn't expected.
* Map/Set: There are a couple issues to work through here:
    * Object-typed keys are supposed to match by identity, not value. However, serialization loses identity. So an object-typed key becomes sort of nonsensical. Should we disallow them? (Note that true "structured clone" is a bit different here since it actually does retain identity of objects serialized as part of the same message, although TBH it still seems super-confusing to rely on that.)
    * Should we support promise pipelining on maps and sets?
    * If so, should we also support `.map()` on maps and sets?
* ArrayBuffer: If we support this, I think we probably also want to support all the typed array types? But typed arrays are pretty specific to JavaScript, and I wonder if this creates pain for implementations in, say, Python.
* URL & Headers: These ones seem like no-brainers to support.

lmaccherone:
I hand wrote the description, but I do ask it questions, when I'm writing. I'm pretty sure, this line was a cut and paste of its response to me asking it, "Did we follow all Cap'n Web coding patterns?" The individual encoders/decoders (the name for them in my [Lumenize](https://lumenize.com) project) were ported for this PR but I was using a marker approach and you use a tupple approach so we had to adapt them and I wanted to make sure we did it the Cap'n Web way instead of the Lumenize way. LLMs are good at answering questions like that.:
```
All changes follow existing Cap'n Web patterns. Type detection uses prototype matching (or instanceof for
URL/Headers). Serialization/deserialization follows the existing ["type", data] array format. Recursive types 
(Map, Set) use the same 64-level depth limit as objects and arrays.
```
BTW, I back ported your tupple approach to my [`@lumenize/structured-clone`](https://www.npmjs.com/package/@lumenize/structured-clone) package. I haven't released that version yet, but thanks for the idea. There was a slim but not zero risk of content collision with our marker approach and the tupple approach fixes that and it turns out that it is more effiicient also.

I use AI coding now almost all the time. I hand write all the doc-tests ([example doc-test about Cap'n Web vs Lumenize RPC](https://github.com/lumenize/lumenize/blob/main/doc-test/rpc/capn-web-comparison-just-works/test/just-works.test.ts)) and I review carefully every line of test code the AI writes. I don't review every line of code anymore but that's a recent evolution of my process.
 
As for the types, I understand your thinking to a degree but I thought that parity with Workers RPC would be beneficial. Otherwise, "it just works" isn't really true. I was targeting that parity with the exception of cycles/aliases which you indicated you want to remove from Workers RPC at some point. 

My structured-clone handles cycles/aliases and it's a capability we rely upon heavily because our primary data structure is a DAG with lots of aliases. Yes, we could have refactored to use an edges and nodes structure (and we may someday) but in the meantime, we're stuck with sending that over the wire.

The special numbers group is just how we did it on Lumenize so that's how we ported it.

I was just trying to do the right thing in contributing rather than complaining about the lack of type support in Cap'n Web, but understand if you can't use any of this. Let me know if you want me to update any of this or if you cherry pick any of it out to use.

I'll give 0.2.0 of Cap'n Web a spin when I update my comparison to Lumenize RPC in the next few weeks. Great stuff. Love it.

lmaccherone:
I was thinking about your pushback on Map/Set

To me, progress beats perfection — never let perfect be the enemy of better.

Yeah, I hear you on Map/Set identity keys. But I’d still rather have useful Map/Set support that covers 90% of cases than hold it back because of imperfectly faithful identity. Few will notice the keys are different on the other end of the wire. The ones that do, will already understand that they couldn't be the same on the other end?

Re: .map() — honestly, that’s where things start feeling too clever. Clarity beats cleverness. (Martin Fowler’s been saying that for years — he’s right.) It’s one of those features that looks cool on paper but adds more mental overhead than it's worth. Just add a method on the other end that does the transformation and call that.

Few will discover .map() in Cap'n Web and most that do won't trust it. The ones that put in the effort to understand it enough to trust it may still reject it as too clever. When I first saw that in Cap'n Web and read how you do it, I was impressed, but I still don't think I'll ever put it in my own code.

The one thing I think it must have though is what you call promise pipelining. TBH, I don't quite follow that terminology. My own RPC uses something I call operation chaining and nesting (OCAN). You can build a complex chain/batch/nesting structure of thenable Proxy object, sometimes with the same one being referenced more than once. When it "sees" the `await`/`.then` it converts all that to a single OCAN that goes over the wire. To OCAN it doesn't matter what the type of the parameters are as long as you can encode them as a plain javascript object.

lmaccherone:
Here's my primer on OCAN if you are interested: https://lumenize.com/docs/rpc/operation-chaining-and-nesting

lmaccherone:
I haven't published this to lumenize.com yet and won't for at least a few days, but here's a writeup that my AI coding agent and I did this morning on [Set/Map behavior in Lumenize RPC](https://github.com/lumenize/lumenize/blob/main/website/docs/structured-clone/maps-and-sets.mdx). It's closer to Workers RPC than Cap'n Web.

Note, this PR does not go nearly as far at preserving identity across that boudary. It'll essentially only work as expected for primative Map keys or Set values. I still think that's valuable, but cries out for documentation.

This comment and the one above was completley hand written. No LLM check before posting. No LLM help in writing except for asking for the exact Martin Fowler quote.

kentonv:
When I mentioned promise pipelining on maps and sets, what I meant was, can I write:

```
stub.getSomeMap().get("key").foo();
```

That is, `getSomeMap()` returns an `RpcPromise<Map<string, Foo>>`, and I can immediately enqueue a pipelined operation on a specific element of the map.

I actually don't think we necessarily have to support this! It's a bit weird in that you have to know, in advance, that the map will contain some particular key. With maps, you usually don't know that -- if they keys are known in advance, you're usually using plain objects instead. So we could say "yeah Map is supported but you can't pipeline on it". I believe Workers RPC implements it this way, in fact.

In fact, I'd argue that supporting `.map()` is much more useful than supporting pipelining on an individual key. It's the same for arrays -- you _can_ pipeline on an array element, but it's awkward to do so, since you usually don't know the size of the array until it has actually returned to you, so how do you know which indexes are valid? Usually, though, what you really want is to apply the same operation on every element -- hence, `.map()`. I think the same logic applies to `Map`. But I'd also be fine with not supporting it initially. But we should _probably_ reserve the right to implement it later, which means we have to _disable_ the default `.map()` behavior that applies to non-arrays, where it checks for nullishness and then applies to the value if non-null.

(Obviously, I disagree that `.map()` is too clever, but I don't expect we'll convince each other on that point.)

lmaccherone:
As close to Workers RPC parity is what I was going for.

What's the main use case for .map()? Is it when you don't have control over the code you are calling? I just can't imagine why I wouldn't prefer adding a method to do the collection manipulation over .map() from remote?

burdiyan:
I think `.map()` is the best thing about Cap'n Web, and is probably _the_ thing that makes it super valuable. It allows you to get rid of N+1 problem and other waterfall requests that are common in remote APIs.

lmaccherone:
Fair enough, but you don't have the N+1 problem either if you create a remote method?

lmaccherone:
Don't get me wrong. It's cool. I'm just trying to understand under what conditions I would favor it over the new method approach?

kentonv:
I think this thread is the wrong place to debate the utility of `.map()` but if someone wants to start a discussion feel free. :)

dimitropoulos:
@lmaccherone I'm eager get this stuff over the line, so eager that I want to tackle these things individually to get them moved along faster (since, there's a non-overlapping set of things to discuss for Map/Set that don't apply to RegExp, etc.).  So I'm closing since there hasn't been any activity in many months, and here's my read on what's next:

- ArrayBuffer`, `DataView`, and typed arrays were added in https://github.com/cloudflare/capnweb/pull/201
- ["inf"], ["-inf"], ["nan"] were added for the special-numbers (as discussed above)
- I split out the Map/Set things into https://github.com/cloudflare/capnweb/issues/221, we'll handle separately
- I split out the RegExp part into https://github.com/cloudflare/capnweb/issues/222, and opened a PR: https://github.com/cloudflare/capnweb/pull/225
- support for Headers landed in https://github.com/cloudflare/capnweb/pull/135
- I split out the URL support part into https://github.com/cloudflare/capnweb/issues/223 and opened a PR: https://github.com/cloudflare/capnweb/pull/224 crediting your initial work

lmaccherone:
Thanks @dimitropoulos! Glad someone else wants this.

That said, I've moved on. When I submitted this PR, I had refactored my own [@lumenize/structured-clone](https://lumenize.com/docs/structured-clone/) to mimic Cap'n Web's arrays-as-tuples format, so it was essentially an extension of the Cap'n Web serialization format. I was hoping to switch my own use to Cap'n Web's implementation when the PR was accepted. 

However, @kentonv saying that he was never going to support cycles/aliases and even planned to remove it from Workers RPC scared me because I depend on that. That, combined with the pushback and delay here for the other stuff, had me convinced months ago to just move on.

Since then, I've completely refactored the format away from the arrays-as-tuples Cap'n Web approach to use an approach like SuperJSON because I did [some experiments](https://github.com/lumenize/lumenize/blob/main/experiments/structured-clone-object-format/RESULTS.md) on raw size and, more importantly, JSON merge patch size so I could send partial updates over the wire. The format I ended up with was 21% more compact than the Cap'n Web-like one for the raw object itself, and a single-element patch required only 64-113 bytes using JSON merge patch. 

I like JSON merge patch for a number of reasons, including that it's a standard but also that it's baked into SQLite.

lmaccherone:
Honestly, I feel like Cap'n Web and RpcTarget are a dead end now. Besides this type-support struggle, the juice just isn't worth the squeeze for session management. The most valuable things sessions buys — the [capability trust model](https://lumenize.com/docs/mesh/calls#operation-chaining) and the ability for the [output of one call to be the input to another in a single round trip](https://lumenize.com/docs/mesh/calls#operation-nesting) — are better achieved using continuation serialization in my @lumenize/mesh package. I get the exact same `lmz.call()` API all the way to/from browser clients with the exact same rich type support as I do between DOs, WorkerEntrypoints, DO facets, etc. and I never need to worry about a dropped session.

dimitropoulos:
cool cool - thanks @lmaccherone! `@lumenize/mesh` looks cool, and I'm always a supporter of exploring where the capability trust model can go!  stay in touch if you have any breakthroughs you find valuable to share!

PR #100, opened 2025-11-03
setup some release stuff

threepointone:
Mostly copied from workers-oauth-provider. I still need to setup repo secrets and a couple of apps, but opening this rn.  

EDIT: added secrets and apps. 

---

This sets up some stuff for release automation, built around github actions, changesets, and pkg.pr.new. 
- pkg-pr-new.yml: For every pull request, we publish a version to their registry, and a comment will popup in the PR comments showing how to use it (eg: https://github.com/cloudflare/workers-oauth-provider/pull/99#issuecomment-3454787239) This lets people test out PRs before they even land. 
- prerelease.yml: whenever something lands in main, after Test(test.yml) passes, we publish to npm with `@beta`, with the version being the git hash. This lets people test whatever's in master quickly.  
- release.yml: When a PR lands after Test(test.yml) passes, the changeset action opens a Version Packages PR (if not already created), that deletes all changesets under the .changeset folder, and populates the CHANGELOG with their contents. You can edit the CHANGELOG further as you wish here. 
- release.yml: When the Version Packages PR itself lands, then the changesets action will do an npm release (with the highest common changeset level: path, minor, or major). It will also tag git and make a github release. 

You can make a changeset in 2 ways:
- running `npx changeset` in your PR branch, which will ask you a couple of questions (what level of change, and title). You can add more details directly in the generated changeset. You can add as many (or as few, even 0) changesets as you'd like. 
- Every PR will have a comment generated that shows what changesets are included (or if they're not). This will also have a link to generate and add a changeset to the PR directly (example https://github.com/cloudflare/workers-oauth-provider/pull/99#issuecomment-3454786218)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,13 +25,8 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: 22
+          node-version: 24    # includes npm 11.5.1 or later, needed for trusted publishing
           cache: "npm"
-
-      # Ensure npm 11.5.1 or later is installed, to use trusted publishing.
-      # See: https://docs.npmjs.com/trusted-publishers
-      - name: Update npm
-        run: npm install -g npm@latest
 
       - run: npm ci
       - run: npm run build
@@ -61,13 +56,8 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: 22
+          node-version: 24    # includes npm 11.5.1 or later, needed for trusted publishing
           cache: "npm"
-
-      # Ensure npm 11.5.1 or later is installed, to use trusted publishing.
-      # See: https://docs.npmjs.com/trusted-publishers
-      - name: Update npm
-        run: npm install -g npm@latest
 
       - run: npm ci
 


.github/workflows/release.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - id: changesets
+        uses: changesets/action@v1
+        with:
+          version: npx tsx .github/changeset-version.ts
+          publish: npx tsx .github/changeset-publish.ts
+        env:
+          GITHUB_TOKEN: ${{ secrets.CAPNWEB_GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
+          NPM_PUBLISH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
```
kentonv: Maybe we can use the new trusted publishing stuff?

https://github.blog/changelog/2025-07-31-npm-trusted-publishing-with-oidc-is-generally-available/

Then we have no secrets, which feels nice. No secrets = nothing to leak.

.github/workflows/release.yml
```diff
@@ -0,0 +1,80 @@
...
+    environment: release
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
```
threepointone: same here

kentonv (approved):

PR #107, opened 2025-11-05
Add CommonJS build output and update dependencies

threepointone:
- generate both ESM and CommonJS outputs and adjust package exports
- upgraded some devDependencies
- removed the sample changeset I'd added
- update workflow to not explicitly call build (since we do that with prepublish anyway)

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,11 +11,15 @@
   ],
   "exports": {
     ".": {
+      "types": "./dist/index.d.ts",
       "import": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -11,36 +11,40 @@
   ],
   "exports": {
     ".": {
+      "types": "./dist/index.d.ts",
```
threepointone: this needs to be first in the list

kentonv (approved):

PR #124, opened 2025-12-31
fix(types): assignability error in tests stub

dimitropoulos:
I happened to notice there's a few errors in [`index.test.ts`](https://github.com/cloudflare/capnweb/blob/main/__tests__/index.test.ts).  Seems like that's because the build skips the tests (naturally) and vitest doesn't do typechecking. 

here's the error:
```
Argument of type '(i: number) => Promise<number>' is not assignable to parameter of type 'Unstubify<((i: Unstubify<number>) => Promise<number> & { toString: (radix?: Unstubify<number | undefined>) => Promise<string> & { [x: number]: Promise<...> & ... 2 more ... & StubBase<...>; ... 50 more ...; toWellFormed: () => Promise<...> & ... 2 more ... & StubBase<...>; } & { ...; } & StubBase<...>; ... 4 more .....'.
  Type '(i: number) => Promise<number>' is not assignable to type '(i: number) => number'.
    Type 'Promise<number>' is not assignable to type 'number'.ts(2345)
```

image

### DIFF
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -673,7 +673,7 @@ describe("basic rpc", () => {
     expect(await stub.jsonify({x: 123, $remove$toJSON: () => "bad"})).toBe('{"x":123}');
   });
 
-  it("supports passing async functinos", async () => {
+  it("supports passing async functions", async () => {
     await using harness = new TestHarness(new TestTarget());
 
     async function square(i: number) {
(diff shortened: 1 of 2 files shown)


dimitropoulos:

kentonv:
Thanks!

FWIW, I think there are also a couple places in the test where TypeScript complains about infinite recursion, which seems like a deeper problem with the way the types work right now that needs to be addressed.

`types.d.ts` actually originates as a copy of [workerd's `rpc.d.ts`](https://github.com/cloudflare/workerd/blob/main/types/defines/rpc.d.ts), but with several refinements (that should probably be backported to workerd at some point). However, it is still a bit janky. Could definitely use some help from a TypeScript expert to get this all working properly!

kentonv (approved):

PR #132, opened 2026-02-06
Add support for sending ReadableStream and WritableStream over RPC, with automatic flow control.

kentonv:
When you send a WritableStream over RPC, the remote side gets a WritableStream. They can write to it. If they write faster than the connection can handle, or faster than your app actually consumes the data, they'll experience backpressure.

When you send a ReadableStream over RPC, the RPC system immediately begins reading from the stream and sending the chunks over the wire so that they are already ready for the remote end to read when it starts reading. This again applies backpressure appropriately. Under the hood, we ask the other end to create a "pipe" -- exposing a WritableStream back to us -- and then we pump chunks into that stream. Meanwhile, the call receiver receives the read end of the pipe.

The flow control is at present based on a fixed window size of 256kb per WritableStream. I intend to fix that in a subsequent change but this is enough to unblock remote bindings that use streams.

### DIFF
--- a/src/rpc.ts
+++ b/src/rpc.ts
@@ -538,7 +538,7 @@ class RpcSessionImpl implements Importer, Exporter {
     return readable;
   }
 
-  createPipe(readable: ReadableStream): ImportId {
+  createPipe(readable: ReadableStream, readableHook: StubHook): ImportId {
     if (this.abortReason) throw this.abortReason;
 
     this.send(["pipe"]);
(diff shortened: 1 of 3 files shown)


kentonv:
Update: I went ahead and added adaptive window size adjustment!

dmmulroy (commented):
Two bugs found during review. Failing tests and fixes on separate branches (both parented to this PR's HEAD).

PR #133, opened 2026-02-08
Add support for WireFormat + binaryBytes passthrough

ashkalor:
- Introduces a WireFormat interface so that users can plug in binary formats (e.g. CBOR)
- Widens all transports (RpcTransport, WebSocket, MessagePort, batch) from `string` to `string | ArrayBuffer`
- Adds binaryBytes option to pass Uint8Array through raw instead of base64-encoding
- Backwards compatible as jsonFormat is set as default. 

My main motivation was to be able to directly pass messages without having to message -> base64 -> stringify  - > parse -> decode -> message. 
If we use a package like CBOR, message -> encode -> decode (which is very fast) -> message, skipping the whole base64 encode and decode. This basically allows someone in the userland to define their own interface that works with RPC

```ts
  import { type WireFormat, newWebSocketRpcSession } from "capnweb";
  import { encode, decode } from "cbor2";

  const cborFormat: WireFormat = {
    encode(value: unknown): ArrayBuffer {
      return encode(value);
    },
    decode(data: string | ArrayBuffer): unknown {
      if (typeof data === "string") {
        throw new TypeError("cborFormat expects binary data, got string");
      }
      return decode(new Uint8Array(data));
    },
  };

  // Both sides must use the same format
  let stub = newWebSocketRpcSession<MyApi>("wss://example.com/rpc", undefined, {
    format: cborFormat,
    binaryBytes: true,  // pass Uint8Array raw instead of base64-encoding
  });

  let result = await stub.processImage(new Uint8Array(pixelData));
```

I did some benchmarks for this 

### Wire Size

| Scenario | JSON | CBOR | Savings |
|---|---:|---:|---:|
| Small RPC call (~40B) | 31 B | 22 B | 29% |
| Medium object (~2KB) | 1.3 KB | 916 B | 30% |
| Large binary (64KB) | 85.4 KB | 64.0 KB | 25% |
| OccupancyGrid (1024x1024) | 1.3 MB | 1.0 MB | 25% |
| PointCloud (50K pts, 800KB) | 1.0 MB | 781.4 KB | 25% |

### Performance

| Scenario | Op | JSON | CBOR | Winner |
|---|---|---:|---:|---|
| Small RPC (~40B) | encode | 125 ns | 3.28 µs | JSON 26x |
| | decode | 234 ns | 2.33 µs | JSON 10x |
| | full RT | 984 ns | 6.41 µs | JSON 7x |
| Medium obj (~2KB) | encode | 2.04 µs | 193.56 µs | JSON 95x |
| | decode | 4.82 µs | 35.62 µs | JSON 7x |
| | full RT | 11.40 µs | 238.23 µs | JSON 21x |
| Large binary (64KB) | encode | 9.59 µs | 29.85 µs | JSON 3x |
| | decode | 12.14 µs | 2.48 µs | **CBOR 5x** |
| | full RT | 44.04 µs | 36.57 µs | **CBOR 1.2x** |
| OccupancyGrid (1M) | encode | 152.82 µs | 369.45 µs | JSON 2x |
| | decode | 194.68 µs | 10.13 µs | **CBOR 19x** |
| | full RT | 783.11 µs | 395.57 µs | **CBOR 2x** |
| PointCloud (800KB) | encode | 121.17 µs | 245.80 µs | JSON 2x |
| | decode | 139.23 µs | 6.45 µs | **CBOR 22x** |
| | full RT | 554.20 µs | 255.84 µs | **CBOR 2x** |

P.S: Huge fan of your work btw!  Been following your work from capnproto days!

### DIFF
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -5,7 +5,7 @@
 import { expect, it, describe, inject } from "vitest"
 import { deserialize, serialize, RpcSession, type RpcSessionOptions, RpcTransport, RpcTarget,
          RpcStub, newWebSocketRpcSession, newMessagePortRpcSession,
-         newHttpBatchRpcSession} from "../src/index.js"
+         newHttpBatchRpcSession, type WireFormat, jsonFormat} from "../src/index.js"
 import { Counter, TestTarget } from "./test-util.js";
 
 let SERIALIZE_TEST_CASES: Record<string, unknown> = {
@@ -111,15 +111,17 @@ class TestTransport implements RpcTransport {
     }
   }
 
-  private queue: string[] = [];
+  private queue: (string | ArrayBuffer)[] = [];
   private waiter?: () => void;
   private aborter?: (err: any) => void;
   public log = false;
 
-  async send(message: string): Promise<void> {
+  async send(message: string | ArrayBuffer): Promise<void> {
     // HACK: If the string "$remove$" appears in the message, remove it. This is used in some
     //   tests to hack the RPC protocol.
-    message = message.replaceAll("$remove$", "");
+    if (typeof message === "string") {
+      message = message.replaceAll("$remove$", "");
+    }
 
     if (this.log) console.log(`${this.name}: ${message}`);
     this.partner!.queue.push(message);
@@ -130,7 +132,7 @@ class TestTransport implements RpcTransport {
     }
   }
 
-  async receive(): Promise<string> {
+  async receive(): Promise<string | ArrayBuffer> {
     if (this.queue.length == 0) {
       await new Promise<void>((resolve, reject) => {
         this.waiter = resolve;
@@ -161,11 +163,11 @@ class TestHarness<T extends RpcTarget> {
 
   stub: RpcStub<T>;
 
-  constructor(target: T, serverOptions?: RpcSessionOptions) {
+  constructor(target: T, serverOptions?: RpcSessionOptions, clientOptions?: RpcSessionOptions) {
     this.clientTransport = new TestTransport("client");
     this.serverTransport = new TestTransport("server", this.clientTransport);
 
-    this.client = new RpcSession<T>(this.clientTransport);
+    this.client = new RpcSession<T>(this.clientTransport, undefined, clientOptions);
 
     // TODO: If I remove `<undefined>` here, I get a TypeScript error about the instantiation being
     //   excessively deep and possibly infinite. Why? `<undefined>` is supposed to be the default.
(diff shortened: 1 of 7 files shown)


ashkalor:

ashkalor:
Hey any updates on this? @kentonv If this approach is against the vision of this library, I'd be happy to make necessary changes to accomodate the vision. Please let me know.

kentonv:
Hi, sorry, this is in my inbox but it's a pretty busy inbox and I tend to handle the capnweb stuff in batches every couple weeks.

Looking briefly, I wonder if a different design would work with less code needed: Instead of giving the system a `WireFormat` hook, what if you could declare that your transport wants to send and receive messages in JS object format rather than string format. In this case the system would pass the pre-encoded objects to `send()` and expect already-decoded objects from `receive()`. The transport can then do whatever encoding it wants internally.

Actually I could imagine we have multiple levels of encoding:
* string: JSON encoding applied already.
* JSON-ready: JS object, but all contents are JSON-compatible.
* JSON-ready + Uint8Array: JS object except bytes are kept in Uint8Array format, for encoders that support that.
* Structured clone: JS object that is structured clonable. The system only applies encoding to RPC stubs. We could change the MessagePort transport to use this.

ashkalor:
That makes a lot of sense, so basically you want this to be handled in the transport layer. Are you comfortable with me adding Generic type support to RPC transport? Something like below 

```ts
 // Core types in src/serialize.ts:
 export type EncodingLevel = "stringify" | "devalue" | "partial" | "passthrough";
 export type TransportData<L extends EncodingLevel> = L extends "stringify" ? string : object;

 // Generic RpcTransport in src/rpc.ts:
 export interface RpcTransport<L extends EncodingLevel = "stringify"> {
   readonly encodingLevel?: L;
   send(message: TransportData<L>): Promise<void>;
   receive(): Promise<TransportData<L>>;
   abort?(reason: any): void;
 }

```
Where 
stringify -> json 
devalue -> plain js object 
partial -> Partial devalue + Uint8array 
passthrough -> structured clone. 


Also do you want me to close this pull request and create a new one for these changes? Or can I force push the new changes directly under the same branch?

SamJB123:
In case it helps at all, I've had stable consistent results for the last month or so in personal projects using cbor-x's record system for automatic dynamic schema optimisation. 

The aim was to preserve the "it just works" functionality while making the most of the persistent websocket connections. But with simple binary passthrough to support yjs etc as well. I was surprised by how stable it's been.

Ask Claude/codex to read over the implementation here if you're curious. Feel free to pillage. There's a bunch of benchmarks/tests written in there too. https://github.com/SamJB123/capnweb

(Holding out for hibernation support to really roll it out, that would be an amazing day, literally just ended up back here yearning for signs of that. I have some awesome distributed agent system capability-passing working with this, nothing but pure RPC messages, single entry point per user across a complex mesh..complex reactive authorisation logic etc... but that also means I'm massively impacted by hibernation vs not! Have had to fall back to using CF agents websocket RPC in the meantime.)

ashkalor:
Closing this in favour of #144 .

PR #136, opened 2026-02-13
fix(types): add callback contextual typing (#134)

### LINKED ISSUE #134: Callback parameter type not inferred from RpcStub (implicitly any)
Been testing out this library -- awesome work. I noticed when calling an RPC method that accepts an RpcStub with a callback, the callback parameter type is not inferred and defaults to any. When explicitly adding the type, it works fine, and also (correctly) gives you a type error if you put the wrong type. But when you leave it empty, it defaults to any.

See the example below for what I mean:

```ts
import { newWebSocketRpcSession, RpcStub, RpcTarget } from 'capnweb'

class MyRpcServer extends RpcTarget {
    sendMeCurrentTs(callback: RpcStub<(ts: number) => void>) {
        callback(Date.now());
    }
}

const stub = newWebSocketRpcSession<MyRpcServer>('ws://example.com');

stub.sendMeCurrentTs((ts) => { // Parameter 'ts' implicitly has an 'any' type. Why isn't it being inferred as number?
    console.log('received ts:', ts);
});

// BELOW WORKS AS EXPECTED

stub.sendMeCurrentTs((ts: string) => { // type error (as expected, ts should be a number)
    console.log('received ts:', ts);
});

stub.sendMeCurrentTs((ts: number) => { // no type issues, works as expected
    console.log('received ts:', ts);
});
```

I do not believe this is the same problem as #112 since I am using `"lib": ["ESNext"]` in my tsconfig. I also tested and did not have the issue mentioned there.

VastBlast:
Fixes #134 

As far as I'm aware, this should not cause any regressions in existing typing.

### DIFF
--- a/src/types.d.ts
+++ b/src/types.d.ts
@@ -108,7 +108,9 @@ export type Stubify<T> =
 // `Stub` depends on `Provider`, which depends on `Unstubify`, which would depend on `Stub`.
 // prettier-ignore
 type UnstubifyInner<T> =
-  T extends StubBase<infer V> ? (T | V)  // can provide either stub or local RpcTarget
+  // Keep local RpcTarget acceptance, but avoid `Stub | Value` unions when the stub
+  // is already assignable to the value type (needed for callback contextual typing).
+  T extends StubBase<infer V> ? (T extends V ? V : (T | V))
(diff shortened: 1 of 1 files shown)


VastBlast:

VastBlast:
Superseded by #142

PR #138, opened 2026-02-18
Move build format option to tsup config

VastBlast:
Simple PR to move the formatting option to the tsup config. It doesn't make sense to hardcode `--format esm,cjs` twice in package.json which overrides the tsup config. It should be added in the tsup config directly. That hardcoded command was added in #107 [around here](https://github.com/cloudflare/capnweb/pull/107/changes#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R30).

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,8 +27,8 @@
     "access": "public"
   },
   "scripts": {
-    "build": "tsup --format esm,cjs",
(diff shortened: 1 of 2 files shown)


kentonv:
Thanks!

PR #139, opened 2026-02-18
fix: Handle cloudflare-internal:workers default export

penalosa:
When running as a workerd extension module (like miniflare's `dispatch-namespace.worker.ts`), only `cloudflare-internal:workers` is available, rather than `cloudflare:workers`. It's easy enough for Miniflare to rewrite the capnpweb imports to `cloudflare-internal:workers` when bundling it's internal workers, but the `import *` approach causes problems, since the internal module exports everything on `default` rather than as named exports:

```javascript
// cloudflare:workers (regular workers)
import * as cfw from "cloudflare:workers";
cfw.RpcTarget // ✅ works

// cloudflare-internal:workers (extension modules)
import * as cfw from "cloudflare-internal:workers";
cfw.RpcTarget // ❌ undefined
cfw.default.RpcTarget // ✅ works
```

This change normalizes `workersModule` to handle both cases by checking for a `default` export first.

### DIFF
--- a/.changeset/pink-teachers-sneeze.md
+++ b/.changeset/pink-teachers-sneeze.md
@@ -1,0 +1,5 @@
+---
+"capnweb": patch
+---
+
+fix: Handle `cloudflare-internal:workers` default export
--- a/src/core.ts
+++ b/src/core.ts
@@ -28,6 +28,8 @@ if (!Promise.withResolvers) {
 }
 
 let workersModule: any = (globalThis as any)[WORKERS_MODULE_SYMBOL];
+// Handle cloudflare-internal:workers which exports on `default` instead of named exports
+workersModule = workersModule?.default ?? workersModule;
 
 export interface RpcTarget {
   [__RPC_TARGET_BRAND]: never;


penalosa:

kentonv (changes requested):
I'm uncomfortable with this fix as it means that if `cloudflare:workers` would never be able to add a `default` export without breaking Cap'n Web users in the wild. We cannot force-update Cap'n Web so we would not be able to fix it on the Cap'n Web side.

Moreover, it's unclear that this is a lasting fix as there could be further differences between `cloudflare-internal:workers` and `cloudflare:workers` that may only reveal themselves in the future when Cap'n Web tries to use more of the `cloudflare:workers` interface.

IMO the problem here is a problem with workerd extension modules, they should be allowed to import `cloudflare:workers`. This seems like an important thing to fix in general as it is increasingly the case that you need to import `cloudflare:workers` to properly implement a lot of things.

penalosa:
@kentonv fair enough—I'll close this and try and see if I can open a PR to workerd

PR #142, opened 2026-02-20
fix(types): rewrite type definitions

### LINKED ISSUE #53: "Type instantiation is excessively deep and possibly infinite" when calling a method on an untyped session

### LINKED ISSUE #54: Returning a bare `Promise` from map errors at runtime without a TypeScript error
Reproduction:

```ts
class MyApiServer extends RpcTarget {
  getNames() {
    return ["Alice", "Bob", "Charlie"];
  }

  greet(name: string) { return `Hello ${name}` }
}

declare const api: RpcStub<MyApiServer>

await api.getNames().map(async name => {
  await someTask();
  return api.greet(name);
})
```

IMO this should be a type error to be able to call `.map()` type safe. The user should get an error here because they're returning something that's not `Result` or valid container or constant serializable value. I'd expect when returning a native promise to get an error here.

### LINKED ISSUE #118: `unknown` inference is broken
I generally disallow explicit any in code bases I work on and prefer `unknown` over `any` as it enforces type guards.

However, with capnweb specifying `unknown` as the return type for any method results in `never` return types when using RpcStub. Initially discovered as we had a `Record<string, unknown>` in a nested type (which is also broken).

### LINKED ISSUE #134: Callback parameter type not inferred from RpcStub (implicitly any)

VastBlast:
This PR rewrites type definitions to fix a ton of issues. I believe this closes all the open & confirmed type issues/bugs reported. Based on my testing, I did not find any new issues, but please let me know if anything slipped.

My motivation for this stemmed from the current type definitions being incredibly hard to work with in my current project due to the bugs. The library is very neat in terms of actual functionality and performance (which I love) but the types are hard to work with.

I will admit, AI assisted with a lot of this *(gpt-5.3-codex:Xhigh)*, but I did ultimately rewrite parts myself. However, my methodology was not blindly using AI: I first wrote several type tests and created a type-testing suite to run against many different base/edge cases. I then tested it against the current types, before using AI to both rewrite the types. After many hours, feedback loops, and handwritten rewrites, I finally landed on this.

I decided against adding those tests to this PR since I do not know if that is something wanted or structured in a specific way. Let me know if you'd like them added and I can push them here.

While working on this, I noticed [a comment left](https://github.com/cloudflare/capnweb/issues/54#issuecomment-3348826748) by @kentonv:
> Thanks but getting the types right is a pretty complicated issue and I think it's best left to the team that understands the model precisely.

So I understand if this is being worked on internally and you'd rather not accept a PR, but I do think its worth looking into these types and potentially using what you can from it.

### Here are the list of changes/fixes:
- `any` return values no longer leak through as `any`.
  - `parse(): any` now gives `api.parse(): Promise<unknown>` (not `Promise<any>`). 
  > Fixes #53
- Inline callback params are contextually inferred from `RpcStub<(...args) => ...>` signatures.
  - with `run(cb: RpcStub<(n: number) => void>)`, `stub.run(n => n.toFixed())` infers `n` as `number` (not `any`).
  > Fixes #134 
- `map()` callback results now reject native promises (including nested ones) and inputs are intentionally non-thenable placeholders.
  - `api.listUsers().map(u => Promise.resolve(u.getName()))` is a type error.
  - `api.listNames().map(name => name.then(...))` is now a type error.
  > Fixes #54
- `unknown` return values are now handled consistently.
  - `getUnknown(): unknown` gives `api.getUnknown(): Promise<unknown>` instead of degrading to unusable/`never` shapes.
  > Fixes #118
- `unknown` object fields are now accepted in RPC-compatible types.
  - `Record<string, unknown>` is treated as valid `RpcCompatible`.
  > Fixes #118
- `map()` callback placeholder values can now be forwarded into params directly.
  - `api.listIds().map(id => api.getUser(id))` type-checks even though `id` is a map placeholder, not a plain number.
- Callable placeholders in `map()` keep their call signatures. 
  - if each item is `(n: number) => string`, then `formatter(1)` is allowed and `formatter("x")` is rejected.
- Array vs tuple providers are now typed separately for better index behavior. 
  - for `Promise<readonly [number, User]>`, `stub[0]` is typed from the tuple slot instead of generic array fallback.
- `Stubify`/`Unstubify` now filter to string/number keys in mapped objects, so symbol-keyed properties are excluded from RPC object mapping.
- `UnstubifyAll` now uses `readonly unknown[]`, so for example methods expecting `readonly [string, number]` keep that in stub call args.

### DIFF
--- a/.changeset/tidy-hounds-brush.md
+++ b/.changeset/tidy-hounds-brush.md
@@ -1,0 +1,5 @@
+---
+"capnweb": minor
+---
+
+Major improvements to type definitions, fixing bugs and making them more accurate.


kentonv:
Thanks, these look like sorely needed improvements to the types!

I will need to find some time to review this carefully -- it may take a week or two as I'm pretty swamped at the moment. But I wanted to let you know I've seen this and am eager to integrate it as bizarre type errors are a sore spot I'm seeing a lot of right now.

vicary:
My honest suggestion is to add type-only tests with recursive nesting, and a N-depth type inference.

We are manually restricting the type depth in our codebase by policy. So it's quite useful for us to know the new depth limit, if this gets merged.

kentonv:
Hi @VastBlast!

Thanks so much for this PR. These fixes look great and I think are more thorough than what I would have likely come up with.

A couple requests:
* I'd love to have the type tests. Could you add them to the PR?
* It looks like you re-ordered RpcCompatible, StubBase, and Stub relative to TypedArray and BaseType, which makes the diff harder to read. Could you move them back to their old ordering to make the diff clear?

Thanks!

kentonv:
And sorry again for the slow review, it's been an excessively busy few weeks.

VastBlast:
> Hi @VastBlast!
> 
> Thanks so much for this PR. These fixes look great and I think are more thorough than what I would have likely come up with.
> 
> A couple requests:
> 
> * I'd love to have the type tests. Could you add them to the PR?
> * It looks like you re-ordered RpcCompatible, StubBase, and Stub relative to TypedArray and BaseType, which makes the diff harder to read. Could you move them back to their old ordering to make the diff clear?
> 
> Thanks!

Thank you for taking the time to review this. I restored the original order and made the diff easier to read, as requested.

I also added tests (*note: gpt-5.3-codex:Xhigh assisted in writing a lot of these tests*). But as I was doing last minute checks, I noticed another existing bug that hasn't been fixed by this PR yet. I was initially going to fix it, but I am unsure if this is a typing or runtime bug, based on a comment I read. I opened a new issue with details: #146

If this is in fact a typing bug, I can try to fix it. However, based on a quick assessment it might cause typing issues for cf workers unless we include an ugly/hacky or partial solution which isn't ideal. But let me know if #146 is indeed a typing bug and I'll look further into it.

kentonv:
That's a missing feature. `Promise<T>` ought to be supported and work similarly to `RpcPromise<T>`, but I haven't gotten around to implementing it yet.

I don't think you should try to implement it in this PR.

While in theory it would be best to make the types match the implementation in the meantime, I think this would be difficult. We would have to explicitly recognize `RpcPromise` and support it, while rejecting `Promise`, which will make the types a lot more complicated. So I'm fine with leaving them as-is for now and eventually fixing the runtime to match.

VastBlast:
> That's a missing feature. `Promise<T>` ought to be supported and work similarly to `RpcPromise<T>`, but I haven't gotten around to implementing it yet.
> 
> I don't think you should try to implement it in this PR.
> 
> While in theory it would be best to make the types match the implementation in the meantime, I think this would be difficult. We would have to explicitly recognize `RpcPromise` and support it, while rejecting `Promise`, which will make the types a lot more complicated. So I'm fine with leaving them as-is for now and eventually fixing the runtime to match.

Ah, I see. And yes it would complicate stuff. Since it’s a planned feature, I agree it would be better to keep it as is rather than switch back and forth.

kentonv:
Thanks so much! You saved me a ton of work here.

PR #144, opened 2026-03-04
feat: add encoding levels to RPC transport

ashkalor:
Implements the encoding level architecture suggested by @kentonv in #133, handling serialization at the transport layer rather than through a global WireFormat hook.

### Summary

- Add `EncodingLevel` type with four levels: `stringify`, `devalue`, `partial`, `passthrough`
- Add optional `encodingLevel` property to `RpcTransport` interface
- Add `wrapTransport()` helper for custom binary formats (CBOR, MessagePack, etc.)
- Serialize/deserialize based on transport's encoding level
- Maintain full backwards compatibility (`stringify` is the default)

### Encoding Levels

| Level | Format | Use Case |
|-------|--------|----------|
| `stringify` | JSON string | Default, max compatibility |
| `devalue` | JS objects | Avoid double-parse overhead |
| `partial` | JS objects + raw Uint8Array | Binary data without base64 |
| `passthrough` | Structured-clonable | postMessage, Workers |

### Motivation

Binary formats like CBOR can significantly outperform JSON for large payloads by avoiding base64 encoding cycles. This architecture lets transports declare their capabilities and have serialization handled appropriately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -692,27 +692,36 @@ Note that you should not use a `Window` object itself as a port for RPC -- you s
 
 ### Custom transports
 
-You can implement a custom RPC transport across any bidirectional stream. To do so, implement the interface `RpcTransport`, which is defined as follows:
+You can implement a custom RPC transport across any bidirectional stream. For string-based transports, implement `RpcTransport`:
 
 ```ts
-// Interface for an RPC transport, which is a simple bidirectional message stream.
 export interface RpcTransport {
-  // Sends a message to the other end.
-  send(message: string): Promise<void>;
+  // Sends a JSON string message. Returns the byte size of the message.
+  // Transport errors should be propagated via receive() rejecting.
+  send(message: string): number;
 
   // Receives a message sent by the other end.
-  //
-  // If and when the transport becomes disconnected, this will reject. The thrown error will be
-  // propagated to all outstanding calls and future calls on any stubs associated with the session.
-  // If there are no outstanding calls (and none are made in the future), then the error does not
-  // propagate anywhere -- this is considered a "clean" shutdown.
   receive(): Promise<string>;
 
-  // Indicates that the RPC system has suffered an error that prevents the session from continuing.
-  // The transport should ideally try to send any queued messages if it can, and then close the
-  // connection. (It's not strictly necessary to deliver queued messages, but the last message sent
-  // before abort() is called is often an "abort" message, which communicates the error to the
-  // peer, so if that is dropped, the peer may have less information about what happened.)
+  // Called when the RPC system needs to abort the session.
+  abort?(reason: any): void;
+}
+```
+
+For transports with custom encoding (CBOR, MessagePack, structured clone, etc.), implement `RpcTransportWithCustomEncoding`:
+
+```ts
+export interface RpcTransportWithCustomEncoding {
+  // Declares what encoding level this transport uses.
+  readonly encodingLevel: "json" | "jsonWithBytes" | "structuredClone";
+
+  // Encodes and sends a message. Returns the encoded byte size if known
+  // (for flow control), or void if unavailable (e.g. structured clone).
+  send(message: unknown): number | void;
+
+  // Receives and decodes a message.
+  receive(): Promise<unknown>;
+
   abort?(reason: any): void;
 }
 ```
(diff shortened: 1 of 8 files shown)


burdiyan:
Having some examples in the README.md would be appreciated 🙏.

ashkalor:
> Having some examples in the README.md would be appreciated 🙏.

Hey just added them, let me know if it looks good. Didn't initially add them because I was expecting some kind of feedback.

README.md
```diff
@@ -736,3 +736,64 @@ let stub: RemoteMainInterface = session.getRemoteMain();
 ```
 
+
+### Encoding Levels
```
kentonv: This section is written more like a change description than documentation. Phrases like "existing code works unchanged" don't really make sense in a README.

I would reduce this to just like:

> By default transports send and receive string messages. However, transports that want more control over the serialization can declare the property `encodingLevel` to tell Cap'n Web what format they accept, with the following options:

Then include the table.

Add this to the previous section ("Custom Transports"), doesn't need to be its own section.

README.md
```diff
@@ -736,3 +736,64 @@ let stub: RemoteMainInterface = session.getRemoteMain();
...
+    return myDecoder.decode(await this.connection.read());
+  }
+}
+```
+
+What happens to `Uint8Array([1, 2, 3])` at each level:
+- `"stringify"` → `'["bytes","AQID"]'` (JSON string)
+- `"devalue"` → `["bytes", "AQID"]` (JS object)
+- `"partial"` → `["bytes", Uint8Array([1,2,3])]` (raw binary)
+- `"passthrough"` → `["bytes", Uint8Array([1,2,3])]` (also preserves Date, BigInt, Error)
```
kentonv: This section is written more like a change description than documentation. Phrases like "existing code works unchanged" don't really make sense in a README. I also think there's too much information here -- this is an obscure feature most readers don't need to know so much about.

Let's just add this to the end of the previous section ("Custom Transports"):

```suggestion

By default, `send()` accepts a string, and `receive()` returns a string, with Cap'n Web handling the encoding all the way to and from strings. However, transports that want more control over the serialization can declare the property `encodingLevel` to control just how much encoding Cap'n Web does before passing off the message:

* `"string"`: The default. Messages are strings.
* `"json"`: Messages are JSON-compatible objects. The transport is responsible for serializing/deserializing.
* `"jsonWithBytes"`: Like "json" except that byte arrays are left as `Uint8Array` instead of base64-encoded. Handy for use with serializations like CBOR or MessagePack that support this efficiently.
* `"structuredClone"`: Messages are structured-clonable objects. Cap'n Web will only implement special handling of RPC stubs. This is useful when the transport is a `MessagePort` or similar.
```

Also let's change the `encodingLevel` names to these, I think they are more understandable.

src/rpc.ts
```diff
@@ -3,17 +3,27 @@
...
+   * - "partial": Transport receives/sends JS objects with raw Uint8Array.
+   * - "passthrough": Transport receives/sends structured-clonable objects.
+   */
+  readonly encodingLevel?: EncodingLevel;
+
   /**
    * Sends a message to the other end.
    */
-  send(message: string): Promise<void>;
+  send(message: string | object): Promise<void>;
```
kentonv: Hmm, it's disappointing that we lose type safety here. It'd be nice to fix that.

One way would be like:

```
export type RpcTransport = {
  encodingLevel?: "string";
  send(message: string): Promise<void>;
  receive(): Promise<string>;
  abort?(reason: any): void;
} | {
  encodingLevel: "json" | "jsonWithBytes" | "structuredClone";
  send(message: unknown): Promise<void>;
  receive(): Promise<unknown>;
  abort?(reason: any): void;
}
```

But that breaks anyone who uses `implements RpcTransport` today, because `RpcTransport` is no longer an `interface`.

Another approach might be:

```
export interface RpcTransport<Level extends EncodingLevel = "string"> {
  readonly encodingLevel: Level;
  
  send(message: Level extends "string" ? string : unknown): Promise<void>;
  receive(): Promise<Level extends "string" ? string : unknown>;
  abort?(reason: any): void;
}
```

But this forces the level to be written twice, and there's no apparent way to make `evcodingLevel` optional in the "string" case, so this still breaks existing users (forcing them to specify `encodingLevel`).

Ugh.

Maybe the best thing is just to declare two different types:

```
export interface RpcTransport {
  readonly encodingLevel?: "string";
  
  send(message: string): Promise<void>;
  receive(): Promise<string>;
  abort?(reason: any): void;
}

export interface RpcTransportWithCustomEncoding {
  readonly encodingLevel: "json" | "jsonWithBytes" | "structuredClone";
  
  send(message: unknown): Promise<void>;
  receive(): Promise<unknown>;
  abort?(reason: any): void;
}
```

And then we just accept `RpcTransport | RpcTransportWithCustomEncoding` in `RpcSession`'s constructor. Ugly but no breakage.

kentonv (commented):
Really like where this is going! Getting the types right seems a bit tricky.

src/rpc.ts
```diff
@@ -607,17 +661,17 @@ class RpcSessionImpl implements Importer, Exporter {
   }
 
   sendStream(id: ImportId, path: PropertyPath, args: RpcPayload)
-      : {promise: Promise<void>, size: number} {
+      : {promise: Promise<void>, size?: number} {
```
kentonv, line 664: I see you changed `size` to optional here, which passes type checks because it ends up being returned by `StubHook.stream()`, where it just so happens that `size` is also declared as optional.

However, `StubHook.stream()`'s contract is that if no `size` is returned, then the call is _local_, and `promise` does not resolve until said local call actually completes. That is not the contract you are providing here. So this breaks streaming in the case that `size` isn't available -- the caller will queue up the entire stream in a tight loop, creating excessive buffer bloat for large streams.

I think what we will need to do here is, if the underlying transport doesn't provide a size, then we need to compute an estimate of the size by walking the message. We only need to do this in `sendStream`, and only when the transport doesn't provide a size.

README.md
```diff
@@ -735,4 +744,33 @@ let stub: RemoteMainInterface = session.getRemoteMain();
 // Now we can call methods on the stub.
 ```
 
-Note that sessions are entirely symmetric: neither side is defined as the "client" nor the "server". Each side can optionally expose a "main interface" to the other. In typical scenarios with a logical client and server, the server exposes a main interface but the client does not.
+Note that sessions are entirely symmetric: neither side is defined as the "client" nor the "server". Each side can optionally expose a "main interface" to the other. In typical scenarios with a logical client and server, the server exposes a main interface but the client does not.ś
```
kentonv, line 747: Seems to be a stray `ś` added to this line.

But also this file has been rewritten in a different way than what I suggested before, could you please follow the suggestion?

kentonv:
I'm happy to take over the PR at this point and finish it up if you prefer -- can be a lot faster than back-and-forth code review.

README.md
ashkalor: Hey I put a comment above to clarify my intent here , nevertheless I ll clean this up as per your suggestion if thats what you want here.

src/rpc.ts
ashkalor: Hey I put a comment above talking about this. The main reason I made this optional was because it needed to sit nice with structuredClone type of encoding. Its not straightforward to estimate the size there so not sure how to go by for it. If we are planning to compute the size by running JSON.stringify or something similar wouldnt that be additional overhead that we were aiming to reduce with this implementation? Just putting out my thoughts here loudly, happy to fix this in whatever way you see fit. The other transports currently do return size maybe we can somehow mandate that while keeping size here optional?

ashkalor:
> I'm happy to take over the PR at this point and finish it up if you prefer -- can be a lot faster than back-and-forth code review.

Hey I'd really like to contribute here, its just that I only get free over the weekend because of work. I can work quickly with you for the next two days if you want to see this through to completion.  This would also be my first official open source contribution.

teamchong:
Thanks @ashkalor 

Closing this in favor of #186

PR #145, opened 2026-03-08
Fix base64 handling and also support Buffer.

### LINKED ISSUE #126: Call stack size exceeded error when sending large UInt8Array
The library seems to have trouble sending large files. The following code throws:
```
Error sending file of size 112640 bytes: RangeError: Maximum call stack size exceeded
```

Repro: https://github.com/hansottowirtz/capnweb-large-file-repro.git

`server.ts`:
```ts
import http from "node:http";
import { WebSocketServer } from 'ws';
import { RpcTarget, newWebSocketRpcSession } from "capnweb";

export class Api extends RpcTarget {
  async getFileSize(file: Uint8Array): Promise<number> {
    return file.length;
  }
}

const httpServer = http.createServer();

const wsServer = new WebSocketServer({ server: httpServer })
wsServer.on('connection', (ws) => {
  newWebSocketRpcSession(ws as any, new Api());
});

httpServer.listen(8000);
```

`client.ts`:
```ts
import type { Api } from "./server";

const session = newWebSocketRpcSession<Api>("ws://localhost:8000");

for (let i = 0; i < 100; i++) {
  const size = 1024 * (10 * i);
  try {
    const file = new Uint8Array(size);
    const result = await session.getFileSize(file);
    console.log(`Sent file of size ${size} bytes, received result ${result}`);
  } catch (error) {
    console.error(`Error sending file of size ${size} bytes:`, error);
    break;
  }
}
```

kentonv:
* base64 encoding of large arrays no longer throws on any platform.
* When toBase64 is unavailable, but Buffer is, this uses Buffer.
* Also, we now accept Buffer for serialization and treat it like Uint8Array.
* Also, when decoding, if Buffer is available, we decode to it. This is convenient for Node users.
* Also fixed the stripping of base64 padding (trailing `=`).

Fixes #126

Opencode+Opus: https://share.opencode.cloudflare.dev/share/VwnZjojz

### DIFF
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -150,6 +150,16 @@ describe("simple serialization", () => {
     expect(new Uint8Array(deserialized)).toStrictEqual(bytes);
   })
 
+  it("strips base64 padding from serialized bytes", () => {
+    // 5 bytes requires base64 padding (5 % 3 != 0), verify it's stripped.
+    let bytes = new Uint8Array([1, 2, 3, 4, 5]);
+    let serialized = serialize(bytes);
+    expect(serialized).not.toContain("=");
+    let deserialized = deserialize(serialized) as Uint8Array;
+    expect(deserialized).toBeInstanceOf(Uint8Array);
+    expect(new Uint8Array(deserialized)).toStrictEqual(bytes);
+  })
+
   it("can serialize Node.js Buffer as bytes", () => {
     if (typeof Buffer === "undefined") return; // skip in browsers
     let buf = Buffer.from("hello!");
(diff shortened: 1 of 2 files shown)


kentonv:
@jasnell I'm curious if you have an opinion on this.

With this change, Cap'n Web will automatically deserialize byte arrays as `Buffer`, if it is available, otherwise it will use `Uint8Array`. `Buffer` is a subclass of `Uint8Array` so this seems pretty safe, and also more convenient on Node (and node_compat mode in Workers, presumably).

Any concerns?

jasnell:
The are some quirky differences with methods like slice but otherwise should be fine

PR #152, opened 2026-03-16
Fix serialization for Invalid/NaN dates

### LINKED ISSUE #150: RPC serialization fails for Invalid/NaN Date
Returning an Invalid Date, `new Date(NaN)`, causes a serialization error because JSON converts NaN values to null. It should be returned as Invalid Date without throwing an error.

Reproduction:
```ts
import { RpcTarget, newMessagePortRpcSession } from "capnweb";

class MyApi extends RpcTarget {
  getInvalidDate(): Date {
    return new Date(NaN);
  }
}

const localApi = new MyApi();
console.log(localApi.getInvalidDate()); // Invalid Date, no error (as expected)

async function main() {
  const channel = new MessageChannel();
  newMessagePortRpcSession(channel.port1, new MyApi());
  const api = newMessagePortRpcSession<MyApi>(channel.port2);

  const invalidDate = await api.getInvalidDate(); // TypeError: unknown special value: ["date",null]
  console.log(invalidDate); // expected: Invalid Date, without throwing an error
}

main();
```

VastBlast:
Fixes #150 to correctly serialize/deserialize invalid/NaN dates

### DIFF
--- a/.changeset/proud-melons-stare.md
+++ b/.changeset/proud-melons-stare.md
@@ -1,0 +1,5 @@
+---
+"capnweb": patch
+---
+
+Fix serialization for Invalid/NaN dates
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -169,6 +169,16 @@ describe("simple serialization", () => {
     expect(deserialized).toBeInstanceOf(Uint8Array);
     expect(new Uint8Array(deserialized)).toStrictEqual(new Uint8Array(buf));
   })
+
+  it("preserves Invalid Date values through serialization", () => {
+    let invalidDate = new Date(NaN);
+    let serialized = serialize(invalidDate);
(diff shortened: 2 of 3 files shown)


kentonv:
Sorry for the very slow review, I've been deep underwater on my inbox for the past month. :(

This seems like a good idea, but I suppose Infinity and -Infinity have the same problem, and we can't encode all of them as `null`. Perhaps we should encode them as `"inf"`, `"-inf"`, and `"nan"`, like we do for the raw numbers?

kentonv:
OK I tried to test this and discovered that `new Date(Infinity).getTime()` actually returns NaN... infinities are not supported by `Date`. Weird!

kentonv:
that's... not the best randomly-generated changeset name but oh well it's a temporary file...

kentonv (approved):

PR #154, opened 2026-03-19
fix: websocket session and callback promise retention leak

### LINKED ISSUE #153: Memory leak on client side with websocket sessions
There seems to be a memory leak on the client side with websocket sessions, unless I am doing something wrong. I understand there are disposal rules to consider, especially with websockets per the docs, but I tried many different things and the only way to clear the memory of unused promises seems to be to restart the session. See the reproduction below.

I also ran it with `--inspect` and it seems there are forever-growing `Promise`'s kept in the memory.

## Server:
```ts
import { RpcTarget, newWebSocketRpcSession } from "capnweb";
import { WebSocketServer } from "ws";

class Api extends RpcTarget {
  sendMeText(text: string): string {
    return `got: ${text}`;
  }
}

const server = new WebSocketServer({ port: 8787, perMessageDeflate: false });

server.on("connection", ws => {
  newWebSocketRpcSession(ws as any, new Api());
});

console.log("ws://localhost:8787");
```

## Client:
```ts

type Api = {
  sendMeText(text: string): string;
};

const session = newWebSocketRpcSession<Api>("ws://localhost:8787");
const text = "abc123".repeat(2000);

let calls = 0;

setInterval(() => {
  console.log(
    `calls=${calls} heapUsed=${(process.memoryUsage().heapUsed / 1024 / 1024).toFixed(1)}MB`
  );
}, 1000);


while (true) {
  await session.sendMeText(text);
  ++calls;
  await new Promise(resolve => setTimeout(resolve, 1)); // delay 1ms
}

// this doesn't prevent a memory leak either, same result:
while(true){
  const res = session.sendMeText(text);
  await res;
  res[Symbol.dispose]();
  ++calls;
  await new Promise(resolve => setTimeout(resolve, 1)); // delay 1ms
}
```

VastBlast:
This PR fixes #153, a client-side memory leak in long-lived RPC sessions.

The session read loop was racing each `transport.receive()` call against a single session-long abort promise, which caused `Promise.race()` to accumulate until the entire session was disposed. This changes `RpcSession` to use a fresh cancellation promise for each pending read.

Edit: I also noticed another issue with callbacks causing a memory leak which isn't specifically reproduced in #153 which has been patched in this PR too.

Essentially if you had something like this where the callback is called regularly:
```ts
subscribeToTicks(cb: RpcStub<(ts: number) => void>) {
    const callback = cb.dup();

    let running = true;
    (async () => {
        while (running) {
            callback(Date.now());
            await new Promise(resolve => setTimeout(resolve, 100));
        }
    })()

    function cleanup() {
        running = false;
        callback[Symbol.dispose]?.();
    }

    callback.onRpcBroken(() => { cleanup(); });
}
```

The client would subscribe and capnweb would keep the callback promises in memory even though the result cannot be used.

### DIFF
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -1308,6 +1308,36 @@ describe("stub disposal over RPC", () => {
   });
 });
 
+describe("callback calls", () => {
+  class CallbackTarget extends RpcTarget {
+    async callMany(callback: RpcStub<(text: string) => unknown>, count: number) {
+      let cb = callback.dup();
+      try {
+        for (let i = 0; i < count; ++i) {
+          cb(`value-${i}`);
+          await pumpMicrotasks();
+        }
+      } finally {
+        cb[Symbol.dispose]();
+      }
+    }
+  }
+
+  it("does not retain ignored synchronous callback results", async () => {
+    await using harness = new TestHarness(new CallbackTarget());
+    await harness.stub.callMany((_text: string) => {}, 5);
+    await pumpMicrotasks();
+    harness.checkAllDisposed();
+  });
+
+  it("does not retain ignored synchronous object callback results", async () => {
+    await using harness = new TestHarness(new CallbackTarget());
+    await harness.stub.callMany((text: string) => ({text}), 5);
+    await pumpMicrotasks();
+    harness.checkAllDisposed();
+  });
+});
+
 describe("e-order", () => {
   it("maintains e-order for concurrent calls on single stub", async () => {
     let callOrder: number[] = [];
--- a/src/rpc.ts
+++ b/src/rpc.ts
@@ -323,7 +323,7 @@ class RpcSessionImpl implements Importer, Exporter {
   private reverseExports: Map<StubHook, ExportId> = new Map();
   private imports: Array<ImportTableEntry> = [];
   private abortReason?: any;
-  private cancelReadLoop: (error: any) => void;
+  private cancelReadLoop?: (error: any) => void;
 
   // We assign positive numbers to imports we initiate, and negative numbers to exports we
   // initiate. So the next import ID is just `imports.length`, but the next export ID needs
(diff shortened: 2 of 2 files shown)


kentonv:
The Promise.race() bug makes sense and the fix seems correct. I honestly didn't know that Promise.race doesn't remove its listeners from all the promises once one of them resolves. That's annoying!

Regarding the second issue, I think this is a bug in your app, not in Cap'n Web:

```ts
            callback(Date.now());
```

You are making a call here, and you are neither awaiting it nor disposing the returned promise. You need to do one or the other.

It is very much an intentional part of the protocol design that if you don't await a call promise, the result is not "pulled", and so never proactively sent from the server. The idea is that you now have a reference to the remote result which you can do promise pipelining on, without actually having to transmit it back. You need to dispose your reference (which is the promise itself) when you are done with it.

You could do something like this to ensure the promise is disposed:

```ts
            using promise = callback(Date.now());
```

Assigning the promise to a `using` causes it to be disposed at the end of the scope. (Equivalent to calling `promise[Symbol.dispose]()`.)

However, this is still problematic, because in this usage pattern, you are ignoring errors. If you ignore errors, then there are all sorts of ways your callback _might not run at all_ without you realizing it. If there was a network error, or a serialization error, or any number of other bugs, these would all just be silently ignored.

For this reason, you REALLY need to attach a `.catch()`:

```ts
            callback(Date.now()).catch(err => console.error(err));
```

This also has the side effect of pulling the callback result, so resolves your memory leak.

In any case, if you can prune this PR back to just fix the Promise.race issue, I'm happy to accept that.

kentonv:
I'm sort of losing my mind over the Promise.race memory leak. It's actually leaking the entire messages!

https://x.com/KentonVarda/status/2044935325554823193

dimitropoulos:
my reading of this is that #168 fixed the memory leak, and the second issue was disposing the returned promise, which is separate - and therefore we can close this issue.  if that's not right please let me know and we can reopen and investigate :)

Issue #32, opened 2025-09-23
custom (de)serializer support


cometkim:
I love that this library provides good platform compatibility and the ability to extend its transport. But the wire format is fixed to JSON.

I wonder if it can provide the ability to customize (de)serializer, or even the entire (De)evaluator implementation as well, for example:

```ts
declare interface Serializer;

newWebSocketRpcSession("wss://example.com/api", {
  serializer: myCustomSerializer,
});
```

To allow compacting or compressing wire messages?

smartboyathome:
This would especially be great for those of us who need to handle binary data as part of our events. JSON can't handle this efficiently, as it requires base64 which increases the size of data by 33%, but [MesaagePack](https://msgpack.org/index.html) can handle binary data without encoding.

kentonv:
Yeah... I think it would make sense to adjust the `RpcTransport` interface so that you can capture the messages before the last JSON stringification step. This would also be useful for the MessagePort transport, which doesn't actually need the values to be stringified (since MessagePort can transmit arbitrary stuctured clonable objects).

That said, I think there's another way we might solve the base64 problem. WebSocket supports both textual and binary message types. What I'd like to do is, when we find a sufficiently large `Uint8Array` in a message, pull it out and send it as a separate binary message. Then send the JSON representing the rest of the structure separately, with some sort of an embedded reference to the binary.

smartboyathome:
That seems overly complex, as now you are introducing more state into the server library in order to track binary data and associate it with JSON. Allowing the use of a custom serializer makes this trivial, and the hard work is done for us. I really don't like the idea of defining yet another format just for this use case.

EqualMa:
For now you can implement a custom `RpcTransport` whose `send` method accepts a json string, parses it and then sends whatever you want. Its `receive` method should receive data of that format and return the serialized string.

cometkim:
You can keep the JSON serializer as the default behavior to keep usage simple.

qwtel:
Are you accepting PRs for this? I have a version of this working [qwtel/capnweb](https://github.com/qwtel/capnweb) and could clean it up.

The design is roughly: High level functions accept an optional `Codec` interface that implements encoding and decoding. The `typeForRpc` function has moved into the Codec interface so that `Uint8Array`, etc. can be declared as `"primitive"` to avoid tagging. Includes implementations for object encoding (to avoid stringification for `postMessage`) and `node:v8` serialization.

kettanaito:
[...]

PR #155, opened 2026-03-19
feat: add Blob serialization support over RPC

G4brym:
## Summary

Adds `Blob` as a first-class serializable type — `Blob` objects can now be passed as RPC call arguments and return values, with MIME type preserved across the wire.

## Wire format

```
["blob", "image/png", ["readable", 5]]
```

Bytes always stream through a pipe. Reading a Blob's bytes is inherently asynchronous, so the uniform pipe path keeps the encoder synchronous and matches the wire semantics of a payload containing a promise.

## E-order

The call message is sent synchronously (send-side e-order preserved). On the receiving end, the Evaluator wraps the pipe stream in an `RpcPromise` that resolves to the Blob; delivery to user code waits for all promises in the payload to resolve, same as any other payload containing a promise.

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -38,8 +38,8 @@ export let RpcTarget = workersModule ? workersModule.RpcTarget : class {};
 export type PropertyPath = (string | number)[];
 
 type TypeForRpc = "unsupported" | "primitive" | "object" | "function" | "array" | "date" |
-    "bigint" | "bytes" | "stub" | "rpc-promise" | "rpc-target" | "rpc-thenable" | "error" |
-    "undefined" | "writable" | "readable" | "headers" | "request" | "response";
+    "bigint" | "bytes" | "blob" | "stub" | "rpc-promise" | "rpc-target" | "rpc-thenable" |
+    "error" | "undefined" | "writable" | "readable" | "headers" | "request" | "response";
 
 const AsyncFunction = (async function () {}).constructor;
 
@@ -47,6 +47,9 @@ const AsyncFunction = (async function () {}).constructor;
 // to accept as "bytes". In browsers, this is undefined, which won't match any prototype.
 let BUFFER_PROTOTYPE: object | undefined =
     typeof Buffer !== "undefined" ? Buffer.prototype : undefined;
+
+// Blob is available in every runtime we support (Node >=18, browsers, workerd).
+const BLOB_PROTOTYPE = Blob.prototype;
 
 export function typeForRpc(value: unknown): TypeForRpc {
   switch (typeof value) {
@@ -111,6 +114,9 @@ export function typeForRpc(value: unknown): TypeForRpc {
 
     case Response.prototype:
       return "response";
+
+    case BLOB_PROTOTYPE:
+      return "blob";
 
     // TODO: All other structured clone types.
 
@@ -947,6 +953,7 @@ export class RpcPayload {
       case "bigint":
       case "date":
       case "bytes":
+      case "blob":
       case "error":
       case "undefined":
         // immutable, no need to copy
@@ -1271,6 +1278,7 @@ export class RpcPayload {
       case "primitive":
       case "bigint":
       case "bytes":
+      case "blob":
       case "date":
       case "error":
       case "undefined":
@@ -1409,6 +1417,7 @@ export class RpcPayload {
       case "primitive":
       case "bigint":
       case "bytes":
+      case "blob":
       case "date":
       case "error":
       case "undefined":
@@ -1566,6 +1575,7 @@ function followPath(value: unknown, parent: object | undefined,
       case "primitive":
       case "bigint":
       case "bytes":
+      case "blob":
       case "date":
       case "error":
       case "headers":
--- a/src/core.ts
+++ b/src/core.ts
@@ -48,9 +48,6 @@ const AsyncFunction = (async function () {}).constructor;
 let BUFFER_PROTOTYPE: object | undefined =
     typeof Buffer !== "undefined" ? Buffer.prototype : undefined;
 
-// Blob is available in every runtime we support (Node >=18, browsers, workerd).
-const BLOB_PROTOTYPE = Blob.prototype;
-
 export function typeForRpc(value: unknown): TypeForRpc {
   switch (typeof value) {
     case "boolean":
@@ -115,7 +112,7 @@ export function typeForRpc(value: unknown): TypeForRpc {
     case Response.prototype:
       return "response";
 
-    case BLOB_PROTOTYPE:
+    case Blob.prototype:
       return "blob";
 
     // TODO: All other structured clone types.
--- a/src/serialize.ts
+++ b/src/serialize.ts
@@ -304,13 +304,13 @@ export class Devaluator {
       }
 
       case "blob": {
-        // Blobs are always streamed through a pipe. Reading a Blob's bytes is inherently async,
-        // so there's no way to preserve send-side e-order regardless of encoding; always using
-        // the pipe keeps the encoder synchronous (matching the wire semantics of a payload
-        // containing a promise).
+        // Blobs are streamed through a pipe. This allows very large blobs to be sent without
+        // causing excessively large individual messages nor blocking other messages in the
+        // meantime.
         //
-        // When devaluating via NULL_EXPORTER (i.e. `serialize()`), createPipe() throws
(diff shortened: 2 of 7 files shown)


G4brym:

kentonv:
Hmm, it looks like you went straight for the ReadableStream approach instead of the Uint8Array approach I recommended.

This approach presents additional complexities. On the receiving end, the RPC system will wait for the blob bytes to arrive before delivering anything to the app. This means the call is delivered out-of-order, technically a violation of e-order. Of course, this is the norm when delivering payloads containing promises, so we could just say "blobs behave like promises, they may delay delivery".

But I'm not totally sure if we have to make that compromise. Reading from a `Blob` is asynchronous, so in theory we could construct a `Blob` on the receiving end and deliver it before all the bytes have arrived, then we wait for the bytes when the app actually tries to read the `Blob`. However, I'm not sure the JS `Blob` API provides any way to _construct_ a `Blob` from a stream, only from already-available in-memory bytes. So perhaps we can't really achieve this, absent some new constructor API? I'm not really sure.

Either way, I do think that for _small_ blobs, it is not worth the overhead for streaming, and we really should still encode it as `Uint8Array`. Probably anything up to 64k at least should not be streamed.

G4brym:
Hey @kentonv, ive updated the pr to encode small blobs as Uint8Array, because encoding calls `blob.arrayBuffer()` and this returns a promise, it means smaller blobs will be out of e-order on the sender

kettanaito:
Hi. Don't want to sound like a killjoy, but wouldn't it make more sense to tackle custom serialization as an API first and then implement whatever missing defaults, like `URL` and `Blob`, on top of that API?

That would also allow the developers to implement whichever serialization they are missing instead of waiting for individual PRs to land for those things. It's #32, basically.

kentonv:
Apologies, my inbox has been so full that I haven't gotten to Cap'n Web in a few weeks. I think I can get back to this tomorrow.

> Hi. Don't want to sound like a killjoy, but wouldn't it make more sense to tackle custom serialization as an API first and then implement whatever missing defaults, like URL and Blob, on top of that API?

No absolutely not. Lettings apps add custom serializers is an enormous security footgun that has been a disaster for e.g. Java serialization.

src/core.ts
```diff
@@ -48,6 +48,11 @@ const AsyncFunction = (async function () {}).constructor;
 let BUFFER_PROTOTYPE: object | undefined =
     typeof Buffer !== "undefined" ? Buffer.prototype : undefined;
 
+// Blob.prototype, or undefined in environments that don't have Blob. Same lazy-init pattern as
```
kentonv: Every runtime has supported Blob for a long time, so we don't need to complicate the code with fallbacks.

src/core.ts
```diff
@@ -630,6 +638,7 @@ async function pullPromise(promise: RpcPromise): Promise<unknown> {
 // RpcPayload
 
 export type LocatedPromise = {parent: object, property: string | number, promise: RpcPromise};
+export type LocatedBlobPromise = {parent: object, property: string | number, promise: Promise<Blob>};
```
kentonv: We shouldn't add a separate promise list for blobs. This is too much core complexity just to support an obscure type.

It should be possible to reuse the RpcPromise list. Follow the same pattern as `fixBrokenRequestBody` in serialize.ts to create an RpcPromise from the Promise<Blob>.

src/serialize.ts
```diff
@@ -46,6 +46,31 @@ class NullExporter implements Exporter {
...
+// Collect all bytes from a ReadableStream and construct a Blob with the given MIME type.
+// Used on the receive side to assemble a Blob from a pipe stream before delivering to user code.
+async function streamToBlob(stream: ReadableStream, type: string): Promise<Blob> {
+  // Use Response.blob() where available (browsers, Workers, Node 18+) for efficiency.
+  if (typeof Response !== "undefined") {
+    let b = await new Response(stream).blob();
+    // Response.blob() may not preserve the original MIME type.
+    return b.type === type ? b : b.slice(0, b.size, type);
+  }
+  // Manual fallback for environments without a global Response (e.g. older Node.js).
```
kentonv: No need for this fallback, we don't need to support ancient Node versions.

src/rpc.ts
```diff
@@ -200,9 +200,27 @@ class RpcImportHook extends StubHook {
     let entry = this.getEntry();
     if (entry.resolution) {
       return entry.resolution.call(path, args);
-    } else {
-      return entry.session.sendCall(entry.importId, path, args);
     }
+
+    // If args contain small blobs, pre-read bytes before sendCall() so they can be encoded inline
```
kentonv: I don't think this is worth it.

I had forgotten that the blob cannot be read on the sending side without async ops. I think that means we have no ability to maintain e-order guarantees when blobs are present. _If_ we had the ability to create a blob on the server side that gets filled in asynchronously, that would be a different story, but at present there's no API for that.

But given that we're _forced_ to asynchronously read the blob content, I guess it actually makes sense to always use streaming for blobs. No "inline" threshold.

kentonv (commented):
This is too much complexity to support a relatively obscure type.

I think we should simplify:
* Don't optimize small blobs after all. It doesn't actually accomplish what I wanted anyway (maintaining e-order) since you have to read the blob content asynchronously on the sending side. So it's not worth it.
* Don't add special handling of blob promises, reuse RpcPromise infrastructure.

I think when simplified this implementation should be almost entirely in serialize.ts, with only small changes (adding some new switch cases) in core.ts, and probably no changes at all in rpc.ts.

G4brym:
Hey @kentonv, i just simplified the pr, can you take a look again?

src/core.ts
```diff
@@ -48,6 +48,9 @@ const AsyncFunction = (async function () {}).constructor;
 let BUFFER_PROTOTYPE: object | undefined =
     typeof Buffer !== "undefined" ? Buffer.prototype : undefined;
 
+// Blob is available in every runtime we support (Node >=18, browsers, workerd).
+const BLOB_PROTOTYPE = Blob.prototype;
```
kentonv: Let's inline this constant into the switch below, consistent with all the other types.

src/serialize.ts
```diff
@@ -292,6 +303,21 @@ export class Devaluator {
...
       }
 
+      case "blob": {
+        // Blobs are always streamed through a pipe. Reading a Blob's bytes is inherently async,
+        // so there's no way to preserve send-side e-order regardless of encoding; always using
+        // the pipe keeps the encoder synchronous (matching the wire semantics of a payload
+        // containing a promise).
+        //
+        // When devaluating via NULL_EXPORTER (i.e. `serialize()`), createPipe() throws
+        // "Cannot create pipes without an RPC session." — same behaviour as streams and stubs.
```
kentonv: Let's replace this AI slop comment.

```suggestion
        // Blobs are streamed through a pipe. This allows very large blobs to be sent without
        // causing excessively large individual messages nor blocking other messages in the
        // meantime.
        //
        // Ideally, small Blobs would be inlined. But, there is no way to read a blob
        // synchronously, and we MUST serialize the message synchronously. Hence, we have no choice
        // but to use streaming even for small blobs.
```

src/serialize.ts
```diff
@@ -455,6 +481,17 @@ function fixBrokenRequestBody(request: Request, body: ReadableStream): RpcPromis
   return new RpcPromise(new PromiseStubHook(promise), []);
 }
 
+// Assemble a Blob from a pipe ReadableStream asynchronously, wrapped in an RpcPromise so it plugs
+// into the existing payload-delivery substitution machinery. Same pattern as
+// fixBrokenRequestBody() above: the caller pushes the returned RpcPromise into the Evaluator's
+// `promises` list, and deliverTo() replaces it with the resolved Blob before user code runs.
```
kentonv:
```suggestion
// Unfortuntaely, even though Blobs can only be read asynchronously, there is no way to create
// a blob backed by an asynchronous source; the bytes MUST all be provided upfront. This
// effectively makes it impossible to manitain e-order when sending Blobs.
//
// As a compromise, we deliver a message as if it contained an RpcPromise that resolves to the
// Blob. This has the effect that the RPC system will wait for the whole Blob to stream in before
// delivering the message -- reusing the existing machinery for handling promises.
```

src/core.ts
G4brym: fixed

kentonv (approved):

Dr-Emann:
This should be in protocol.md, I think.

PR #159, opened 2026-03-26
Add bun support

### LINKED ISSUE #61: Bun Support Missing

aron-cf:
Add Bun ServerWebSocket support to capnweb.

Bun's server-side WebSocket uses callback-based handlers registered on `Bun.serve()` rather than the standard `addEventListener` interface. Passing a Bun `ServerWebSocket` to `newWebSocketRpcSession()` silently fails because the event listeners never fire. (#61)

A new `BunWebSocketTransport` implements the `RpcTransport` interface using the same resolver pattern as the existing WebSocket and MessagePort transports. It exposes `dispatchMessage`, `dispatchClose`, and `dispatchError` methods that Bun's handler callbacks invoke to feed messages into the transport.

Two convenience functions sit on top of the transport. `newBunWebSocketRpcHandler()` returns a complete handler object you can pass straight to `Bun.serve()`, wiring everything up automatically via `ws.data`. `newBunWebSocketRpcSession()` returns the stub and transport for users who need custom handler logic.

The readme now documents both approaches and notes that Bun is a supported runtime.

#### Quick start with the handler helper

```ts
import { RpcTarget, newBunWebSocketRpcHandler } from "capnweb";

class MyApi extends RpcTarget {
  greet(name: string) { return `Hello, ${name}!`; }
}

let rpcHandler = newBunWebSocketRpcHandler(() => new MyApi());

Bun.serve({
  fetch(req, server) {
    if (server.upgrade(req)) return;
    return new Response("Not Found", { status: 404 });
  },
  websocket: rpcHandler,
});
```

#### Escape hatch with manual wiring

```ts
import { newBunWebSocketRpcSession, RpcTarget } from "capnweb";

class MyApi extends RpcTarget {
  greet(name: string) { return `Hello, ${name}!`; }
}

Bun.serve({
  fetch(req, server) {
    let userId = authenticate(req);
    server.upgrade(req, { data: { userId } });
  },
  websocket: {
    open(ws) {
      let { stub, transport } = newBunWebSocketRpcSession(ws, new MyApi());
      ws.data.transport = transport;
    },
    message(ws, msg) { ws.data.transport.dispatchMessage(msg); },
    close(ws, code, reason) { ws.data.transport.dispatchClose(code, reason); },
    error(ws, err) { ws.data.transport.dispatchError(err); },
  },
});
```

#### Testing locally

Run the test suite with `npm test`. The new tests in `__tests__/bun.test.ts` exercise the transport, both convenience functions, and full round-trip calls through a loopback pair without needing a real Bun server.

Unit tests cover message queuing, close and error propagation, abort semantics, buffer-to-string conversion, and end-to-end calls including bidirectional callbacks and error forwarding. All existing tests continue to pass.

Closes #61

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Cap'n Web is a spiritual sibling to [Cap'n Proto](https://capnproto.org) (and is
 * That said, it integrates nicely with TypeScript.
 * Also unlike Cap'n Proto, Cap'n Web's underlying serialization is human-readable. In fact, it's just JSON, with a little pre-/post-processing.
 * It works over HTTP, WebSocket, and postMessage() out-of-the-box, with the ability to extend it to other transports easily.
-* It works in all major browsers, Cloudflare Workers, Node.js, and other modern JavaScript runtimes.
+* It works in all major browsers, Cloudflare Workers, Node.js, Bun, Deno, and other modern JavaScript runtimes.
 The whole thing compresses (minify+gzip) to under 10kB with no dependencies.
 
 Cap'n Web is more expressive than almost every other RPC system, because it implements an object-capability RPC model. That means it:
@@ -627,6 +627,71 @@ Deno.serve(async (req) => {
   }
 
   return new Response("Not Found", { status: 404 });
+});
+```
+
+### HTTP server on Bun
+
+Bun's server-side WebSocket API uses [callback-based handlers](https://bun.sh/docs/runtime/http/websockets) instead of the standard `addEventListener` interface. Cap'n Web provides `newBunWebSocketRpcHandler()` which returns a handler object you can pass directly to `Bun.serve()`.
+
+```ts
+import { RpcTarget, newBunWebSocketRpcHandler, newHttpBatchRpcResponse } from "capnweb";
+
+class MyApiImpl extends RpcTarget implements MyApi {
+  // ... define API, same as above ...
+}
+
+// Create a WebSocket handler that manages RPC sessions automatically.
+// The callback is invoked once per connection to create a fresh API instance.
+let rpcHandler = newBunWebSocketRpcHandler(() => new MyApiImpl());
+
+Bun.serve({
+  async fetch(req, server) {
+    let url = new URL(req.url);
+    if (url.pathname === "/api") {
+      // Upgrade WebSocket requests.
+      if (req.headers.get("upgrade")?.toLowerCase() === "websocket") {
+        if (server.upgrade(req)) return;
+        return new Response("WebSocket upgrade failed", { status: 500 });
+      }
+
+      // Handle HTTP batch requests.
+      let response = await newHttpBatchRpcResponse(req, new MyApiImpl());
+      response.headers.set("Access-Control-Allow-Origin", "*");
+      return response;
+    }
+
+    return new Response("Not Found", { status: 404 });
+  },
+
+  // Pass the handler directly — no manual wiring needed.
+  websocket: rpcHandler,
+});
+```
+
+```ts
+import { newBunWebSocketRpcSession, newHttpBatchRpcResponse, type RpcTarget, type BunWebSocketTransport } from "capnweb";
+
+class MyApiImpl extends RpcTarget implements MyApi {
+  // ... define API, same as above ...
+}
+
+type Data = { userId: string; transport?: BunWebSocketTransport<Data> };
+
+Bun.serve<Data>({
+  fetch(req, server) {
+    let userId = authenticate(req);
+    server.upgrade(req, { data: { userId } });
+  },
+  websocket: {
+    open(ws) {
+      let { stub, transport } = newBunWebSocketRpcSession(ws, new MyApiImpl());
+      ws.data.transport = transport;
+    },
+    message(ws, msg) { ws.data.transport?.dispatchMessage(msg); },
+    close(ws, code, reason) { ws.data.transport?.dispatchClose(code, reason); },
+    error(ws, err) { ws.data.transport?.dispatchError(err); },
+  },
 });
 ```
 
--- a/README.md
+++ b/README.md
@@ -669,32 +669,6 @@ Bun.serve({
 });
 ```
 
-```ts
-import { newBunWebSocketRpcSession, newHttpBatchRpcResponse, type RpcTarget, type BunWebSocketTransport } from "capnweb";
-
-class MyApiImpl extends RpcTarget implements MyApi {
-  // ... define API, same as above ...
-}
-
-type Data = { userId: string; transport?: BunWebSocketTransport<Data> };
-
-Bun.serve<Data>({
-  fetch(req, server) {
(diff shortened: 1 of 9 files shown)


aron-cf:

aron-cf:
recheck

kentonv:
In package.json we currently have a special separate build for workerd that includes the workers-only stuff. Maybe we should have a separate build for bun, too? Avoid code bloat for other runtimes.

```
  "exports": {
    ".": {
      "types": "./dist/index.d.ts",
      "import": {
        "workerd": "./dist/index-workers.js",
        "default": "./dist/index.js"
      },
      "require": {
        "workerd": "./dist/index-workers.cjs",
        "default": "./dist/index.cjs"
      }
    }
  },
```

vitest.config.ts
```diff
@@ -17,7 +17,7 @@ export default defineConfig({
           name: 'node',
           // We throw flow-control test under Node only because it's testing straightforward
           // JavaScript -- no need to run it on every runtime.
-          include: ['__tests__/index.test.ts', '__tests__/flow-control.test.ts'],
+          include: ['__tests__/index.test.ts', '__tests__/flow-control.test.ts', '__tests__/bun.test.ts'],
```
kentonv: I'm a bit confused how the bun test passes on node, unless it's not actually testing integration with Bun's builtin APIs. If it isn't testing integration then I don't think it's a useful test.

Can we add Bun to the test matrix, and write a test that actually tests against Bun's built-in APIs?

aron-cf:
@kentonv thanks for the review.

> In package.json we currently have a special separate build for workerd that includes the workers-only stuff. Maybe we should have a separate build for bun, too? Avoid code bloat for other runtimes.

Yeah that makes sense. I've done that.

> I'm a bit confused how the bun test passes on node, unless it's not actually testing integration with Bun's builtin APIs. If it isn't testing integration then I don't think it's a useful test.

Yeh, that's fair. It was testing against the interface of the `BunWebSocket` server. I agree an integration test is more useful here. I've updated the tests to run against an actual Bun webserver.

> Can we add Bun to the test matrix, and write a test that actually tests against Bun's built-in APIs?

I've not created a test matrix to keep things simple, but happy to do so if you'd prefer. Now we now have a `npm run test:bun` script to run the integration tests against the Bun web server and a new `npm run test:ci` to run both `vitest` and `bun` on CI. I've added a note to the README to cover this.

README.md
```diff
@@ -630,6 +630,73 @@ Deno.serve(async (req) => {
 });
 ```
 
+### HTTP server on Bun
+
+Bun's server-side WebSocket API uses [callback-based handlers](https://bun.sh/docs/runtime/http/websockets) instead of the standard `addEventListener` interface. Cap'n Web provides `newBunWebSocketRpcHandler()` which returns a handler object you can pass directly to `Bun.serve()`.
+
+> **Note:** The Bun-specific helpers (`newBunWebSocketRpcHandler`, `newBunWebSocketRpcSession`, and `BunWebSocketTransport`) are shipped in a separate build entry point that Bun resolves automatically via the `"bun"` [conditional export](https://nodejs.org/api/packages.html#conditional-exports) in `package.json`. Other runtimes never load this code, keeping bundle size minimal.
```
kentonv: I think we can leave out this line, it probably isn't worth the cognitive overhead for people to be aware of this.

README.md
```diff
@@ -630,6 +630,73 @@ Deno.serve(async (req) => {
...
+
+    return new Response("Not Found", { status: 404 });
+  },
+
+  // Pass the handler directly — no manual wiring needed.
+  websocket: rpcHandler,
+});
+```
+
+If you need to attach custom data to connections or add your own logic to the WebSocket handlers, use `newBunWebSocketRpcSession()` instead, which gives you direct access to the transport:
```
kentonv: Let's leave this out (here to the end of the section). It can be documented in doc comments in the code, but I don't think we want to spend prime README space on it.

src/bun.ts
```diff
@@ -0,0 +1,140 @@
+// Copyright (c) 2025 Cloudflare, Inc.
```
kentonv: It's 2026. :)

src/index-bun.ts
```diff
@@ -0,0 +1,34 @@
...
+export * from "./index.js";
+
+import { RpcTransport, RpcSessionOptions } from "./rpc.js";
+import { RpcCompatible } from "./types.js";
+import { newBunWebSocketRpcSession as newBunWebSocketRpcSessionImpl,
+         newBunWebSocketRpcHandler,
+         BunWebSocketTransport } from "./bun.js";
+
+export { newBunWebSocketRpcHandler, BunWebSocketTransport };
+export type { BunServerWebSocket } from "./bun.js";
```
kentonv: I don't think we should export this type -- people should use the official types from `@types/bun`, ours is just a minimal subset.

kentonv: Is it possible for us to use `@types/bun` here without forcing non-bun dependents to depend on it? Some way to declare dependencies that only apply to one platform?

README.md
```diff
@@ -654,6 +721,8 @@ function newWebSocketRpcSession(
     : Disposable;
 ```
 
+Note: Bun's `ServerWebSocket` does not implement the standard `WebSocket` `addEventListener` interface. If you are using Bun, use the dedicated `newBunWebSocketRpcHandler()` or `newBunWebSocketRpcSession()` described above.
```
kentonv: Remove this line, it's redundant.

README.md
```diff
@@ -735,4 +804,25 @@ let stub: RemoteMainInterface = session.getRemoteMain();
 // Now we can call methods on the stub.
 ```
 
+Cap'n Web's built-in transports (WebSocket, MessagePort, HTTP batch, and Bun ServerWebSocket) are all implemented on top of this interface.
```
kentonv: Remove. Don't add unrelated stuff to the readme.

README.md
```diff
@@ -735,4 +804,25 @@ let stub: RemoteMainInterface = session.getRemoteMain();
...
+
 Note that sessions are entirely symmetric: neither side is defined as the "client" nor the "server". Each side can optionally expose a "main interface" to the other. In typical scenarios with a logical client and server, the server exposes a main interface but the client does not.
+
+## Development
```
kentonv: Remove new section.

src/index-bun.ts
aron-cf: Hmm, it was only exported for use in the tests. I've pushed an alternative that adds `@types/bun` to `devDependencies`. This means only maintainers depend on it as it's provided by the Bun runtime so feels like a reasonable trade-off.

README.md
kentonv: Looks like you only removed this line, not the rest of the section as requested.

kentonv (approved):

PR #162, opened 2026-04-19
Generalize record-replay: serialize closures passed to any RPC method

ryanrasti:
`<stub>.map(fn)` already records `fn` locally, emits a `["remap", …]` instruction, and replays it on the receiver. This generalizes the mechanism so any user-defined method on an `RpcTarget` can accept callbacks that serialize the same way.

**Motivation**: I'm building [typegres](https://github.com/ryanrasti/typegres) which uses capnweb to expose a capability-based query builder:
```ts
users
    .where(({ users }) => users.id["="](1n))
    .select(({ users }) => ({ id: users.id, name: users.name.upper() }))
    .execute();
```

The callbacks should be interpreted server-side (not exported as a stub and sent back to client).

**Key design decisions**

1. **Backward compatibility**: Plain functions passed over RPC today are exported as stubs (pass-by-reference). To keep that unchanged, the draft only routes functions through record-replay when they're passed inside a .map() callback — where creating a new stub is already forbidden. (No new opt-in flags needed).
2. **Sharing between `map` and general closures**: I chose to re-use the most code possible but maintain distinct wire formats for `remap` and `closure`.

Potential follow-ons (not in this PR):
1. **Sync replay for closures** -- the receiver's replay fn currently returns a Promise because `RpcPayload.deliverResolve` always wraps in one, even when nothing needs awaiting. Async closures are already rejected at serialize time.
  - *Why this matters*: Typegres's query builder is synchronous -- without sync replay, every callback-taking method has to become async, infecting the call chain.
2. **Stub identity preservation**: independent of closures: round-tripping a stub back to its owner currently yields a proxy, not the raw target.  
  - *Why this matters*: server may need to do things like validate the type of the object, access non-exposed fields, etc. which are impossible to do if its own object is wrapped behind a stub

(Opening as a draft to get your read on the design -- happy to rework anything.)

### DIFF
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -1054,6 +1054,92 @@ describe("map() over RPC", () => {
   });
 });
 
+describe("record-replay closure over RPC", () => {
+  it("passes a function as a closure inside a map() callback", async () => {
+    await using harness = new TestHarness(new TestTarget());
+    let counter = new RpcStub(new Counter(0));
+
+    expect(await harness.stub.map(stub => {
+      return stub.callFunction(y => counter.increment(y), 3);
+    })).toStrictEqual({result: 3});
+  });
+
+  it("encodes a closure's literal return value as its terminal instruction", async () => {
+    await using harness = new TestHarness(new TestTarget());
+    let counter = new RpcStub(new Counter(0));
+
+    expect(await harness.stub.map(stub => {
+      return stub.callFunction(y => {
+        counter.increment(y);
+        return 42;
+      }, 3);
+    })).toStrictEqual({result: 42});
+    expect(await counter.value).toBe(3);
+  });
+
+  it("supports invoking a received closure multiple times", async () => {
+    class DoubleCaller extends RpcTarget {
+      async callTwice(fn: RpcStub<(x: number) => Promise<number>>) {
+        return [await fn(1), await fn(2)];
+      }
+    }
+
+    await using harness = new TestHarness(new DoubleCaller());
+    let counter = new RpcStub(new Counter(10));
+
+    expect(await harness.stub.map(stub => {
+      return stub.callTwice(y => counter.increment(y));
+    })).toStrictEqual([11, 13]);
+  });
+
+  it("supports nested closures", async () => {
+    class Passthrough extends RpcTarget {
+      call(fn: RpcStub<(x: number) => number | Promise<number>>, x: number) {
+        return fn(x);
+      }
+    }
+
+    await using harness = new TestHarness(new Passthrough());
+    let counter = new RpcStub(new Counter(0));
+
+    expect(await harness.stub.map(stub => {
+      return stub.call(y => {
+        return stub.call(z => {
+          counter.increment(y);
+          return counter.increment(z);
+        }, 7);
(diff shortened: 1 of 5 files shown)


ryanrasti:

kentonv:
Hmm. Restricting this to only work inside a `.map()` is an interesting way to sidestep the question of which calls _should_ accept pass-by-value closures. But could you give me some more complete examples of how you see this looking in practice? Your example in the description doesn't include the surrounding `.map()`.

ryanrasti:
> Restricting this to only work inside a .map() is an interesting way to sidestep the question of which calls should accept pass-by-value closures. But could you give me some more complete examples of how you see this looking in practice? Your example in the description doesn't include the surrounding .map().

Sure thing. And yes `.map()` is the boundary that activates closure serialization. It's potentially a nice fit because the body of map is pass-by-value by definition:

```typescript
// Table definitions:
class Post extends RpcTarget {
  get id() { return this.column('id') }
  get userId() { return this.column('user_id') }
  get content() { return this.column('content') }
}

class User extends RpcTarget {
  get id() { return this.column('id') }
  posts() { return Post.from().where(({posts}) => posts.userId['='](this.id)) }
}

// Main stub:
class Api extends RpcTarget {
  userForToken(token: string) {
    const userId = validateToken(token)
    return User.from().where(({users}) => users.id['='](userId))
  }

  db(): Database { ... }
}

// Actual usage:
const getNumPosts = (api: RpcStub<Api>) => {
  const token = tokenFromContext(...)
  const results = api.map(_ =>
    api
      .userForToken(token)
      .select(({users}) => ({ 
          id: users.id, 
          numPosts: users.posts().groupBy().select(({posts}) => posts.id.count())
      }))
      .execute(api.db())
  )
  return results[0]?.numPosts ?? 0
}
```

The shape I'm envisioning: wrap the entire query once in a `map()` (or something else that signals pass-by-value), and write the callbacks naturally inside.

ndisidore:
I see you're still pushing here: holler when you want another set of eyes!

dimitropoulos:
hi! are you still interested in proceeding with this?

ryanrasti:
Yes, definitely still interested!

Since opening this PR, I added Sqlite/Durable Object support to Typegres, using a fork of Cap'n Web based on this PR (if interesting: [chat room example](https://github.com/ryanrasti/typegres/tree/main/examples/chat))

This is still a draft because I'm waiting for a read on a couple things that will be necessary:
1. Whether `map()` is the right boundary for pass-by-value closures. The fundamental tension is that closures are pass-by-ref by default and there doesn't seem to be another clean way to annotate "all closures here are pass by value" instead
2. Appetite for two follow-ons my fork introduces (not in this PR, but is in [my fork](https://github.com/ryanrasti/capnweb/commits/efebddf7205b31fbaff8b4ef748d236d876bdeac/)):
  a. Synchronous replay for closures (a small, internal change to not resolve as a promise if there are no internal promises)
  b. Add a `getLocalTarget` (or similar) function: a way to get a reference to the backing local object from a stub (it's a stub because peer passed our own object back to us)

Let me know on the above -- happy to rework to fit the right direction.

PR #166, opened 2026-04-30
Preserve own properties of Error instances across serialization/deserialization

### LINKED ISSUE #87: Custom error details
The protocol spec for the `error` expression is closely tied to JavaScript's `Error` type, and is not very extensible with custom parameters. The spec [says](https://github.com/cloudflare/capnweb/blob/b9a024b7d1881c039358e2e4e0a05bbc27d331fa/protocol.md?plain=1#L113):

> _TODO: We should extend this to encode own properties that have been added to the error._

I wonder what are the plans for supporting the extensibility? Would it be defined as a separate expression type, or just an extension of the existing `error` message?

In my case, I'm experimenting with implementing the server side of Cap'n Web in a different language (Go), and JS-specific error types, besides feeling weird, are also quite limiting — no way to specify error codes, or custom attributes.

I was thinking the extra attributes in JSON and put them into the error message as a string, but it's ugly, and I don't like it.

Maybe one way to support the extensibility could be allowing to use `string | object` as `message` in the `error` expression? This should probably only be supported if `type` is a non-standard JS error type. 

CC: @kentonv.

aron-cf:
Closes #87.

Today the wire format for `Error` only carries `name`, `message`, and (optionally) `stack`. Anything a user attaches to the error — `code`, `details`, a `cause`, the inner errors of an `AggregateError` — is lost the moment the error crosses an RPC boundary.

This PR extends the `error` expression with an optional fifth element, `props`. This uses the same serialization/deserialization as any other object.

`Error.cause` and `AggregateError.errors` are normally non-enumerable, so they are picked up explicitly by the encoder. On the receive side, the decoder constructs the error as before and then assigns each entry from `props` as an own enumerable property on the result.

```ts
let err = new Error("rate limited");
err.code = "E_LIMIT";
err.retryAfter = 30;
throw err;
// receiver sees: err.code === "E_LIMIT", err.retryAfter === 30
```

Properties that cannot be fully serialized are dropped.

Tests have been added to verify the serialization/deserialization of various different error cases and failure modes. There may be too many, happy to remove any if needed.

The `error` expression in protocol.md now documents the optional `props` element, the lossy-fallback semantics.

### DIFF
--- a/.changeset/error-own-properties.md
+++ b/.changeset/error-own-properties.md
@@ -1,0 +1,5 @@
+---
+"capnweb": patch
+---
+
+Errors properties, using `Object.keys()`, are now preserved across the wire. Attach fields like `code` or `details` to an `Error` and they propagate to the other side. The `cause` and `errors` (for `AggregateError`) properties will also be preserved.
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -179,7 +179,7 @@ describe("simple serialization", () => {
     expect(deserialized).toBeInstanceOf(Date);
     expect(Number.isNaN(deserialized.getTime())).toBe(true);
   })
-});
+})
 
 // =======================================================================================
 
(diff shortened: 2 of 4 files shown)


.changeset/error-own-properties.md
```diff
@@ -0,0 +1,19 @@
...
+err.retryAfter = 30;
+throw err;
+// receiver sees: err.code === "E_LIMIT", err.retryAfter === 30
+```
+
+`Error.cause` and `AggregateError.errors` are also preserved, with the inner errors keeping their built-in subclass identity (`TypeError`, `RangeError`, …).
+
+Property values that cannot be represented over the wire (capabilities without a session, cycles, unsupported types) are silently dropped from the error's property bag; the error itself always reaches the receiver. Use `onSendError` to scrub fields explicitly if you need control over what gets sent.
+
+The wire format change is backwards-compatible: errors with no extras are still emitted in the legacy 3- or 4-element form, and older peers that only understand that form will simply ignore the new trailing element.
```
aron-cf: Happy to make this shorter if we want terser changelog entries.

src/serialize.ts
```diff
@@ -297,16 +297,50 @@ export class Devaluator {
...
         }
 
-        let result = ["error", e.name, e.message];
-        if (rewritten && rewritten.stack) {
+        // Capture own enumerable properties plus the standard non-enumerable slots `cause`
+        // and (for AggregateError) `errors`. Each value is run through devaluateImpl so any
+        // supported type round-trips. If a property's value can't be serialized (a capability
+        // without a session, a cycle, an Object.create(null), …) we silently drop it: the
+        // error itself must always make it through. Use `onSendError` to scrub heavy or
+        // sensitive fields explicitly.
```
aron-cf: Would appreciate feedback on whether we just drop the entire property if it fails to serialize rather than making a best effort attempt.

protocol.md
```diff
@@ -180,11 +180,15 @@ A bigint value, represented as a decimal string.
...
 
-_TODO: We should extend this to encode own properties that have been added to the error._
+`props` is an optional fifth element carrying any extra data attached to the error. It is a JSON object whose keys are the error's own enumerable properties (plus the standard non-enumerable `cause` slot, and `errors` for `AggregateError`), and whose values are themselves valid expressions of this protocol round-trip naturally. Property values that cannot be represented are silently dropped from `props`; the error itself always reaches the receiver.
+
+When `props` is present, `stack` is normalised to `null` if absent so that positional indexing for `props` is unambiguous. When there are no extras, the legacy 3- or 4-element form is emitted unchanged.
+
+The `props` element is backwards-compatible: a peer that only understands the 4-element form ignores it and recovers the standard `name`/`message`/`stack`. A peer that understands `props` but receives the 3- or 4-element form simply sees no extras.
```
aron-cf: Happy to remove this if it's too much information but felt it might be useful to have documented.

src/serialize.ts
```diff
@@ -521,10 +555,23 @@ export class Evaluator {
...
             let cls = ERROR_TYPES[value[1]] || Error;
-            let result = new cls(value[2]);
+            // AggregateError's constructor takes (errors, message); we pass an empty array
+            // and patch `errors` from the props bag below.
+            let result = cls === AggregateError ? new cls([], value[2]) : new cls(value[2]);
             if (typeof value[3] === "string") {
               result.stack = value[3];
             }
+            // Optional 5th element: own properties bag. Unknown keys are assigned as own
+            // enumerable properties so the receiver sees what the sender attached.
```
aron-cf, line 630:
```suggestion
            // Optional 5th element: own properties. Unknown keys are assigned as own
            // enumerable properties so the receiver sees what the sender attached.
```

src/serialize.ts
```diff
@@ -521,10 +555,23 @@ export class Evaluator {
         case "error":
           if (value.length >= 3 && typeof value[1] === "string" && typeof value[2] === "string") {
             let cls = ERROR_TYPES[value[1]] || Error;
-            let result = new cls(value[2]);
+            // AggregateError's constructor takes (errors, message); we pass an empty array
+            // and patch `errors` from the props bag below.
```
aron-cf, line 624:
```suggestion
            // AggregateError's constructor takes (errors, message); we pass an empty array
            // and patch `errors` from the props below.
```

src/serialize.ts
```diff
@@ -297,16 +297,50 @@ export class Devaluator {
...
+        if ("cause" in e) {
+          captureProp("cause", anyE.cause);
+        }
+        if (e instanceof AggregateError) {
+          captureProp("errors", e.errors);
+        }
+
+        // Backwards-compat: only emit the new tail elements when there's something to add.
+        // Errors with no extras serialize to the legacy 3- or 4-element form, byte-identical
+        // to what previous versions produced.
```
aron-cf, line 385:
```suggestion
        // Only emit the new tail elements when there's something to add.
        // Errors with no extras serialize to the legacy 3- or 4-element form, byte-identical
        // to what previous versions produced.
```

__tests__/index.test.ts
```diff
@@ -628,6 +821,102 @@ describe("basic rpc", () => {
...
+    expect(caught.request.url).toBe("http://example.com/");
+    expect(caught.request.method).toBe("DELETE");
+  });
+
+  it("silently drops stub-valued error properties without leaking capabilities", async () => {
+    class CounterFactory extends RpcTarget {
+      throwWithStub() {
+        let err = new Error("with-stub") as any;
+        // Attach a stub to the error. There's no sensible lifetime for this capability, so
+        // it must be dropped on the wire rather than leaked to the importer.
```
aron-cf, line 942:
```suggestion
```

.changeset/error-own-properties.md
kentonv: Yeah I think changeset log entries are meant to be one-liners.

src/serialize.ts
```diff
@@ -521,10 +555,23 @@ export class Evaluator {
...
+            // AggregateError's constructor takes (errors, message); we pass an empty array
+            // and patch `errors` from the props bag below.
+            let result = cls === AggregateError ? new cls([], value[2]) : new cls(value[2]);
             if (typeof value[3] === "string") {
               result.stack = value[3];
             }
+            // Optional 5th element: own properties. Unknown keys are assigned as own
+            // enumerable properties so the receiver sees what the sender attached.
+            let props = value[4];
+            if (props && typeof props === "object" && !Array.isArray(props)) {
```
kentonv: If `props` is present but the wrong type, we should presumably throw? (Or actually, `break`, which will lead to the error path.)

src/serialize.ts
kentonv: Dropping the whole property seems reasonable. Trying to encode partial objects seems like it would lead to trouble.

src/serialize.ts
```diff
@@ -297,16 +297,50 @@ export class Devaluator {
...
+        // error itself must always make it through. Use `onSendError` to scrub heavy or
+        // sensitive fields explicitly.
+        let anyE = <any>e;
+        let props: Record<string, unknown> | undefined;
+        let captureProp = (key: string, val: unknown) => {
+          try {
+            let encoded = this.devaluateImpl(val, e, depth + 1);
+            if (!props) props = {};
+            props[key] = encoded;
+          } catch (err) {
```
kentonv: I think we need to do a little more cleanup work here after catching an error. What if it had appended to `exports`? We probably need to unexport the newly-added exports (like we do when catching an error in `devaluate()`).

protocol.md
kentonv: Yeah I think this last paragraph about back-compat is not really worth it, documenting every behavior that existed in a 0.x version is a bit much.

The other two paragraphs should stay.

src/serialize.ts
aron-cf: I think this is now resolved by 7583c778f901 as we drop the property entirely.

src/serialize.ts
aron-cf: Fixed in 7583c778f901, I went with the approach of doing a pre-scan to ensure the property can serialize rather than attempting to serialize then unexporting on error.

protocol.md
aron-cf: Fixed in b56c8b413ce4

src/serialize.ts
aron-cf: Fixed in 49cb3fdcd4aa

.changeset/error-own-properties.md
aron-cf: Fixed in 355659cd77df

aron-cf (commented):
I've reworked the code to skip any error properties that fail to serialize. I've done this in a pre-scan pass (that early exits when it hits a negative) rather than attempting to recover from a serialization error, from what I understand this removes the risk of leaking ReadableStreams at the cost of two passes. Let me know your thoughts.

kentonv (approved):

PR #169, opened 2026-05-11
Add capnweb-validate RPC validators

### LINKED ISSUE #174: runtime type validation
Cap'n Web's biggest weakness right now is that it doesn't validate types at all, but TypeScript can easily lead people to a false sense of security. It's really awkward to write type checks for the types that TS is already assuming are accurate.

I believe the right solution is to create a code generator that, given TypeScript types, generates runtime type-checking code for each type, which can be applied as a wrapper (wrapping each method to check the inputs or returns).

TypeScript famously doesn't want to support this kind of thing, and the port to Go has made it harder. But I think we can do it. One idea is to trick the type checker itself into building the generated code as a string and then outputting it in a diagnostic, which we can then intercept and save to a file.

Note that we would not need to support the entire TS type system – only types that are RPC-compatible ([structured clone](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm)).

teamchong:
Closes #174

## Summary

Adds `capnweb-validate`, an opt-in TypeScript-to-runtime-validator transform for Cap'n Web and Workers RPC services.

Server validation is enabled with `@validateRpc()` on the service class. Users keep importing Cap'n Web APIs from `capnweb`, so the service remains usable with native Workers RPC and with Cap'n Web.

Client validation is enabled by the transform recognizing Cap'n Web session constructors by TypeScript symbol resolution. No client import change is required.

Validated paths:

- server incoming arguments
- server resolved return values
- client outgoing arguments
- client resolved return values

If `@validateRpc()` is left untransformed, it throws a configuration error instead of silently running without validation.

## How it works

- The transform finds `@validateRpc()` decorators imported from `capnweb-validate`.
- It resolves the concrete RPC surface from:
  1. explicit decorator type argument
  2. a single implemented interface
  3. the class public methods
- If a class implements multiple interfaces and no `@validateRpc<T>()` type argument is given, the transform emits a warning and falls back to class public methods.
- It lowers method signatures into runtime validators.
- It emits plain JavaScript validator objects using `capnweb-validate/internal/core`.
- It rewrites the decorator to wrap constructed service instances.
- It recognizes Cap'n Web client session calls from `capnweb` and injects client-side validation wrappers.
- The CLI writes a transformed source tree for Wrangler and other flows without bundler plugin hooks.
- The universal plugin supports Vite, Rollup, Webpack, Rspack, esbuild, and Farm.

Supported pass-by-value types include primitives, arrays, tuples, plain objects, unions, `Record<string, T>`, selected platform built-ins, and `Promise<T>` return unwrapping.

Pass-by-reference values include plain functions, `RpcStub<T>`, `RpcPromise<T>`, `RpcTarget` subclasses, and Workers RPC stubs.

Unsupported wire types such as `Map`, `Set`, `WeakMap`, `WeakSet`, `RegExp`, `ArrayBuffer`, and non-`Uint8Array` typed arrays fail at build time.

## Included

- New `packages/capnweb-validate` workspace package:
  - `@validateRpc()` class decorator
  - `@skipRpcValidation()` method opt-out
  - structured `RpcValidationError`
  - capnweb-free core runtime validators and wrappers
  - Cap'n Web-specific compatibility helpers under `capnweb-validate/capnweb`
  - transform implementation
  - CLI
  - universal bundler plugin adapters
- `capnweb` is an optional peer dependency. The root entry and `capnweb-validate/internal/core` have no `capnweb` imports; Cap'n Web-specific helpers live under `capnweb-validate/capnweb` and `capnweb-validate/internal/capnweb`. Workers RPC-only users can install `capnweb-validate` alone.
- Workers RPC `WorkerEntrypoint` support. Runtime-invoked methods on `WorkerEntrypoint` subclasses (`fetch`, `tail`, `trace`, `scheduled`, `queue`, `test`, `email`, `connect`, `tailStream`) bypass the validator so the Workers runtime can still invoke them; user-defined RPC methods are validated.
- `RpcValidationError extends TypeError` with `err.rpcValidation = { path, expected, actual, value }`. `path` is JSON-pointer-style, for example `["Api", "authenticate", 0, "profile", "email"]`.
- Root build/test wiring.
- Transform/runtime test coverage.
- Client session validation without changing imports from `capnweb`.
- `examples/worker-react` validation wiring.
- VS Code launch/tasks for the example.

## Example

Server:

```ts
import { newWorkersRpcResponse, RpcTarget } from "capnweb";
import { validateRpc } from "capnweb-validate";

type User = { id: string; name: string };

@validateRpc()
class Api extends RpcTarget {
  async authenticate(sessionToken: string): Promise<User> {
    return { id: "u_1", name: "Ada Lovelace" };
  }
}

export default {
  fetch(request: Request, env: Env) {
    return newWorkersRpcResponse(request, new Api());
  },
};
```

Client:

```ts
import type { Api } from "./worker";

const api = newHttpBatchRpcSession<Api>("/api");

await api.authenticate("cookie-123");
```

## Wrangler flow

Wrangler does not expose a bundler plugin hook, so use the CLI for Worker source:

```sh
npx capnweb-validate build --cwd src --out .wrangler/validate
```

Point Wrangler's `main` at the generated entry under `.wrangler/validate`.

The React client uses the Vite plugin, not the CLI.

## Testing

Package:

```sh
npm run -w capnweb-validate build
npm run -w capnweb-validate test
```

Repo:

```sh
npm run build
npm run test:types
```

Example:

```sh
cd examples/worker-react/client && npm run build
node ../../packages/capnweb-validate/dist/cli.cjs build --cwd server --out ../.wrangler/validate
npx tsc -p examples/worker-react/server/tsconfig.json --noEmit
```


kentonv:
Very excited about this!

I probably won't have time to review the code in detail, hopefully @dimitropoulos or @ndisidore can do that.

Going over the design, I have a few questions:

> It resolves the concrete RPC surface from:
> i. explicit decorator type argument
> ii. a single implemented interface
> iii. the class public methods

Is the thinking here that if a class extends an interface, but also has additional public methods, we assume those additional public methods are not intended to be available over RPC?

I see the argument but I'm not sure I agree. We've always made clear that all public string-named prototype methods are available over RPC. If people want local-only methods they need to be symbol-named, but in practice I don't remember ever having a need for this.

> If a class implements multiple interfaces and no `@validateRpc<T>()` type argument is given, the transform emits a warning and falls back to class public methods.

Why can't we emit a validator covering both interfaces?

> It recognizes Cap'n Web client session calls from capnweb and injects client-side validation wrappers.

Hmm, what does this mean exactly? Why does the validator need to interact with Cap'n Web at all? How does this work when using Workers runtime built-in RPC instead of the capnweb module?

My expectation was that if I use `@validateRpc()` on a class, and then I instantiate the class and call the methods locally, those calls will get runtime type checks. That would be totally decoupled from any particular RPC implementation and should "just work" with all of them.

> Workers RPC WorkerEntrypoint support.

How about `DurableObject`? Should work the same.

> RpcValidationError extends TypeError

We might want to just use `TypeError` here as non-standard errors will lose their type when serialized and will end up being just `Error`. I don't really feel like custom errors are that useful anyway.

examples/worker-react/client/vite.config.ts
```diff
@@ -0,0 +1,40 @@
...
+export default defineConfig({
+  plugins: [
+    capnwebValidate.vite({
+      cwd: __dirname,
+      tsconfig: 'tsconfig.json',
+    }),
+  ],
+  resolve: {
+    alias: {
+      'capnweb-validate/internal/capnweb': path.resolve(repoRoot, 'packages/capnweb-validate/src/internal/capnweb.ts'),
```
dimitropoulos: I understand if/why this has to be different from how someone would actually in practice do things (since, we're pointing to a local copy) - can we at least make a comment or something telling people that they won't need to do certain things the same way in their own real-world projects (which would be out of this repo).

packages/capnweb-validate/__tests__/branded-primitive.test.ts
```diff
@@ -0,0 +1,73 @@
+// Branded primitives (`string & { __brand }`) must collapse to their underlying
+// primitive, not be walked as objects (which would reject brand/.toString fields).
+import { describe, it, expect } from "vitest";
+import { transformFixture, transformError } from "./helpers.js";
```
dimitropoulos:
```suggestion
import { transformFixture, transformError } from "./helpers";
```

can we get our tsconfig settings working such that we aren't needing to do the `.js on ts files` dance?  isn't it `bundler` or something? this https://www.typescriptlang.org/tsconfig/#moduleResolution

packages/capnweb-validate/__tests__/helpers.ts
```diff
@@ -0,0 +1,112 @@
+// Shared fixtures for transform tests: build a throwaway project in a tmpdir,
+// run the real transform, and return the generated code plus any warnings.
+import { mkdtempSync, rmSync, writeFileSync, readFileSync } from "node:fs";
```
dimitropoulos: it would be excellent to use a vfs for this.  for example something like https://www.npmjs.com/package/@typescript/vfs (but if typescript is not a dependency, we can use anything).

packages/capnweb-validate/__tests__/transform-module.test.ts
```diff
@@ -0,0 +1,236 @@
...
+  serviceName: string;
+  methods: Record<string, {
+    args: ((value: unknown, path: (string | number)[]) => void)[];
+    returns: (value: unknown, path: (string | number)[]) => void;
+  }>;
+};
+
+async function loadValidator(code: string, binding: string): Promise<TestServiceValidator> {
+  let rt = await import("../src/internal/runtime.js");
+  let prelude = code.match(/import \* as __rt from "capnweb-validate\/internal(?:\/(?:core|capnweb))?";\n([\s\S]*?)\n\s*import\s+/);
```
dimitropoulos:
```suggestion
  let prelude = code.match(/import \* as __cw from "capnweb-validate\/internal(?:\/(?:core|capnweb))?";\n([\s\S]*?)\n\s*import\s+/);
```

can we change this to `__cw` just to make it clear to anyone looking at the outputs (which, I expect people will)

packages/capnweb-validate/__tests__/branded-primitive.test.ts
```diff
@@ -0,0 +1,73 @@
...
+describe("branded primitives", () => {
+  it("validates a symbol-branded string as v.string", () => {
+    const code = emitFor(
+      `type UserId = string & { readonly __brand: unique symbol };
+       class Api extends RpcTarget {
+         async getUser(id: UserId): Promise<string> { return id; }
+       }`
+    );
+    // Assert against the emitted validator only; source (incl. `type UserId`) is kept verbatim.
+    expect(code).toMatch(/getUser[\s\S]*?args:\s*\[\s*__rt\.v\.string\s*\]/);
```
dimitropoulos: instead of doing this kind of regex matching, which seems fragile, is there any way we can just resolve (or eval?) the code so that we can interact with the exports directly, e.g.

```typescript
const __capnweb_validate_Api_server = {
  serviceName: "Api",
  methods: {
    "getUser": { args: [__rt.v.string], returns: __rt.v.string },
  },
};
```

so we can just assert on the actual values like `expect(__capnweb_validate_Api_server).toHaveMethod('getUser')` and stuff like that?

dimitropoulos: perhaps, with that, we can assert on the function names for things like `v.string`

kentonv:
I see some changes were made seemingly in response to my previous comment but I'm not clear on what the answers to my questions are, can you reply directly?

packages/capnweb-validate/__tests__/error-subclass.test.ts
```diff
@@ -0,0 +1,51 @@
...
+      `class Api extends RpcTarget {
+         async a(): Promise<Error> { return null as any; }
+         async b(): Promise<TypeError> { return null as any; }
+       }`
+    );
+    expect(code).toMatch(/"a":\s*\{\s*args:\s*\[\],\s*returns:\s*__rt\.v\.error\s*\}/);
+    expect(code).toMatch(/"b":\s*\{\s*args:\s*\[\],\s*returns:\s*__rt\.v\.error\s*\}/);
+  });
+
+  it("does not hijack a user type merely named Error that is not the global", () => {
```
dimitropoulos, line 62: this is a nice test!  wouldn't have thought to look into this

packages/capnweb-validate/__tests__/generic-class.test.ts
```diff
@@ -0,0 +1,144 @@
...
+    return { code, warns };
+  } catch (e) {
+    return { code: null, warns: [], error: e instanceof Error ? e.message : String(e) };
+  }
+}
+
+describe("generic service classes", () => {
+  it("keeps public class methods outside a single implemented interface", () => {
+    const { code, warns } = compile(
+      `interface A { a(): Promise<string>; }\n@validateRpc()\nclass Api extends RpcTarget implements A { async a(): Promise<string> { return ""; } async extra(x: number): Promise<number> { return x; } }`,
```
dimitropoulos: since we have template strings anyway, it'd be excellent if we could just format these with real newlines so they can be read like normal code

packages/capnweb-validate/__tests__/unsupported-type-hook.test.ts
```diff
@@ -0,0 +1,130 @@
...
+// transports Map/Set via structured clone). Returning "passthrough" emits `any`
+// for that type instead of failing the build; the default stays a build error.
+import { describe, it, expect } from "vitest";
+import { transformFixture, transformError, prelude } from "./helpers.js";
+
+const MAP_SERVICE = `class Api extends RpcTarget {
+  async getConfig(): Promise<Map<string, number>> { return null as any; }
+}`;
+
+describe("onUnsupportedType hook", () => {
```
dimitropoulos: I think the move here is for capnweb-validate to support Map/Set/ArrayBuffer/RegExp (even if it's _before_ capnweb itself supports it).  we can add that support in another PR.

so, effectively, today the validation will pass but capnweb will fail on its own.  that's ok because if we're making this validation library in a way that it's independent of the underlying details, then that's fine.

_THAT SAIDDDDD_, if there's an easy way to add a parameter decorator like `@validateRpc({ exclude: [Map, Set, ArrayBuffer, RegExp]  })` that could be nice.  _even better_ if we can do that at a more global place just once (since, adding this to every `validateRpc` call will get very annoying).  I don't know if we have any way of a runtime-evaluated global way of hooking in at runtime.

packages/capnweb-validate/src/internal/capnweb.ts
```diff
@@ -0,0 +1,155 @@
...
+  type ServiceValidator,
+  type Validator,
+} from "./core.js";
+
+export function __newWorkersRpcResponseWithValidation(
+  request: Request,
+  localMain: object,
+  validator: ServiceValidator
+): Promise<Response> {
+  return capnweb.newWorkersRpcResponse(
```
dimitropoulos, line 27: what about if, instead, we add a place in the upstream `newWorkersRpcResponse` to add validation in the 3rd argument (optionally) like

```ts
capnweb.newWorkersRpcResponse(
  req,
  new Api(),
  {
    validate: __capnweb_validate_Api_server_or_whatever
  }
)
```

while we are trying to mostly stay away from touching the capnweb codebase itself, this seems like a fine tradeoff vs, copying and wrapping capnweb within the capnweb-validate codebase like this (which has the inverse undesired effect of linking capnweb-validate to anything that uses it).

teamchong:
> I see some changes were made seemingly in response to my previous comment but I'm not clear on what the answers to my questions are, can you reply directly?

sorry, I should've replied directly. Answers:

1. Surface resolution: agreed, `implements`/`@validateRpc<T>()` shouldn't be an allowlist. fixed: the surface is now the class's public string-named methods (same as Cap'n Web dispatch). A single implemented interface or an explicit `<T>` is only used to sharpen the signatures of methods that already exist on the class; extra public methods stay exposed and validated.

i.e.
```ts
interface Api {
  greet(name: string): Promise<string>;
}

@validateRpc()
export class Service extends RpcTarget implements Api {
  async greet(name: string): Promise<string> { /* ... */ }

  // Public and string-named -> exposed over RPC and validated, even though
  // it's not in `Api`. Make it private or symbol-named to keep it local.
  status() {
    return "email";
  }
}
```

2. Multiple interfaces: moot after that change. Surface comes from the class, so there's no "which interface wins" case anymore. No warning/fallback path.

3. Cap'n Web coupling: the decorator path is transport-independent. `@validateRpc()` wraps the class instance, so local calls, Cap'n Web server dispatch, and Workers RPC dispatch all hit the same checks.

The Cap'n Web-specific code is just integration around Cap'n Web APIs. Server helpers wrap Cap'n Web boundary functions, and client helpers wrap Cap'n Web stubs because there is no class instance on the client. Workers built-in RPC clients are not wrapped client-side today; the decorated service still validates the server boundary.

4. DurableObject: works the same as WorkerEntrypoint. Added a test: user methods validated, platform hooks (fetch/alarm) pass through.

5. RpcValidationError: switched to plain `TypeError`. (Internally I tag our thrown TypeErrors so warn-mode/union logic can tell ours apart from a user-thrown one, but nothing custom is exported.)

let me know if any of these don't match what you had in mind. still testing the flow before I call it done

kentonv:
> A single implemented interface or an explicit <T> is only used to sharpen the signatures of methods that already exist on the class; extra public methods stay exposed and validated.

Hmm, I think I'd argue that if someone does explicitly write `@validateRpc<T>()`, then we should allow _only_ the methods of T -- other public methods are essentially deemed "invalid" for any call. But if no `<T>` is specified, we should allow all public methods. This makes sense to me as saying "Validate that calls follow exactly the interface T".

> The Cap'n Web-specific code is just integration around Cap'n Web APIs. Server helpers wrap Cap'n Web boundary functions, and client helpers wrap Cap'n Web stubs because there is no class instance on the client. Workers built-in RPC clients are not wrapped client-side today; the decorated service still validates the server boundary.

Hmm, I don't understand what this means. What does it mean for things to be wrapped "client-side"? Why does it make sense for Cap'n Web but not Workers RPC?

kentonv:
Side note: While Cap'n Web's accepted types are similar to structured clone, it's explicitly not exactly the same set. There are some types we support that structured clone doesn't, some we don't support that it does, an notably structured clone supports cycles but we emphatically refuse to.

Just pointing that out since I see a lot of references to "structured clone" here.

teamchong:
> Side note: While Cap'n Web's accepted types are similar to structured clone, it's explicitly not exactly the same set. There are some types we support that structured clone doesn't, some we don't support that it does, an notably structured clone supports cycles but we emphatically refuse to.
> 
> Just pointing that out since I see a lot of references to "structured clone" here.

good point, I’ll stop calling this “structured clone.” The intent here was not to claim Cap’n Web follows structured clone exactly.

Dimitri raised that Workers RPC accepts values like `ArrayBuffer`, `Map`/`Set`, `RegExp`, and typed arrays, and downstream had to change an existing Worker RPC contract from `ArrayBuffer` to `Uint8Array` only because `capnweb-validate` rejected it. So I tried to make the validator target the broader RPC-compatible runtime value set rather than exactly Cap’n Web’s current serializer set.

do you agree with that direction for `capnweb-validate`? In other words: should validation accept Workers-RPC-supported value types and let Cap’n Web transport reject anything it still can’t serialize, or should this package enforce Cap’n Web’s exact transport surface by default?

either way, I’ll rename/remove the “structured clone” wording

kentonv:
Yeah it makes sense to support a superset of what both RPC implementations support, and leave it up to the RPC system to complain if it can't serialize something.

packages/capnweb-validate/__tests__/helpers.ts
teamchong: done. the shared transform fixture now uses `@typescript/vfs` instead of creating temp project files. I kept the real filesystem tests where the filesystem is the behavior under test: CLI/build output, tsconfig discovery, asset copying, symlinks, and outside-cwd handling

packages/capnweb-validate/__tests__/transform-module.test.ts
teamchong: the generated runtime namespace is now `__cw` consistently, including the transform output and test expectations

packages/capnweb-validate/__tests__/unsupported-type-hook.test.ts
teamchong: we now accept `Map` / `Set` / `ArrayBuffer` / `RegExp` in `capnweb-validate`. If `capnweb` itself still doesn’t support one of these at runtime, validation will pass and `capnweb` will fail the same way it would without `capnweb-validate`. So the validator is no longer adding an extra rejection here.

packages/capnweb-validate/__tests__/branded-primitive.test.ts
teamchong: I checked this and the `.js` imports are consistent with the existing Cap’n Web source style, especially for published ESM/TypeScript output. I’m going to leave this PR aligned with the current convention and address extensionless imports separately if we decide to change that repo-wide

examples/worker-react/client/vite.config.ts
teamchong: added a comment above the alias block clarifying that it is only needed for this monorepo example, and external projects using the published packages should not need to copy it

packages/capnweb-validate/__tests__/branded-primitive.test.ts
teamchong: the tests now evaluate the generated validator prelude and assert against the actual validator objects instead of regex-matching emitted source. i.e., branded primitive checks now assert `args[0] === v.string` / `returns === v.string`, and composed validators use runtime behavior or shape introspection. I left source checks only for rewrite mechanics, not validator semantics

packages/capnweb-validate/__tests__/generic-class.test.ts
teamchong: done

teamchong:
> > A single implemented interface or an explicit  is only used to sharpen the signatures of methods that already exist on the class; extra public methods stay exposed and validated.
> 
> Hmm, I think I'd argue that if someone does explicitly write `@validateRpc<T>()`, then we should allow _only_ the methods of T -- other public methods are essentially deemed "invalid" for any call. But if no `<T>` is specified, we should allow all public methods. This makes sense to me as saying "Validate that calls follow exactly the interface T".
> 

pushed a commit to change the explicit type-argument behavior.

```ts
interface Api {
  greet(name: string): Promise<string>;
}

@validateRpc<Api>()
export class Service extends RpcTarget implements Api {
  async greet(name: string): Promise<string> { /* ... */ }

   // this method no in Api
  status() {
    return "email";
  }
}
```

in this case, Api is now the RPC surface. calling status() through the decorated/wrapped instance, including over RPC, will fail with a runtime validation error because status is not in the generated validator.

without the explicit type argument, the surface is still the class public RPC surface, so `status()` remains exposed and validated from the class signature


> > The Cap'n Web-specific code is just integration around Cap'n Web APIs. Server helpers wrap Cap'n Web boundary functions, and client helpers wrap Cap'n Web stubs because there is no class instance on the client. Workers built-in RPC clients are not wrapped client-side today; the decorated service still validates the server boundary.
> 
> Hmm, I don't understand what this means. What does it mean for things to be wrapped "client-side"? Why does it make sense for Cap'n Web but not Workers RPC?

I explained that poorly.

the "client-side wrapping" part is an extra feature for Cap'n Web clients. in Cap'n Web, client code usually creates a remote API object by calling something like `newHttpBatchRpcSession<Api>()`. Since that is a normal function call in user code, the plugin can rewrite it and put a validation wrapper around the returned client stub.

workers RPC client objects are different. the client usually gets them from the platform as env.SERVICE. there is no obvious function call like `newHttpBatchRpcSession<Api>()` for the plugin to rewrite. so this PR validates workers RPC at the decorated server boundary, but does not currently support adding an automatic client-side validation wrapper around Workers RPC bindings.

kentonv:
> in Cap'n Web, client code usually creates a remote API object by calling something like newHttpBatchRpcSession<Api>(). Since that is a normal function call in user code, the plugin can rewrite it and put a validation wrapper around the returned client stub.

Ah so to be clear, this wrapper would validate method _return values_ from `Api`, rather than arguments? (Or if the client passes a stub to the server, then it would validate method arguments on that stub?)

And you're saying capnweb-validate will "rewrite" calls to `newHttpBatchRpcSession<T>()` and similar functions to automatically add validation? How does this rewriting work? The type parameter is obviously needed in order to decide what validation to apply, so runtime monkey-patching can't do the job. I guess you must be using some sort of bundler plugin to do the rewrite?

I think in many cases this client-side validation is not really needed since the client fully trusts the server. So I'm not sure it's desirable to make it automatic. I would say we should give people an explicit way to add the validation, like:

```
let stub = validateStub<T>(newWebSocketRpcSession<T>(...))`
```

Then this explicit `validateStub` function could also be used on Workers RPC stubs.

By the way, we probably need a way to coerce a validated stub to another type, so that e.g. if an API returns `RpcStub<any>`, we can coerce it to a specific `RpcStub<T>`. Note the coercion requires _removing_ the existing validation wrapper and then adding a new wrapper. So this may call for another function like `coerceStub<T>` that does this?

packages/capnweb-validate/src/internal/capnweb.ts
teamchong: I agree this is probably the right long-term shape, but I’d prefer to keep it out of this PR.

adding a third argument to capnweb.`newWorkersRpcResponse()` is a public `capnweb` API change, and I think it deserves its own PR where we can discuss the exact option shape and whether the same hook should exist on the other server entry points too

teamchong:
> > in Cap'n Web, client code usually creates a remote API object by calling something like newHttpBatchRpcSession(). Since that is a normal function call in user code, the plugin can rewrite it and put a validation wrapper around the returned client stub.
> 
> Ah so to be clear, this wrapper would validate method _return values_ from `Api`, rather than arguments? (Or if the client passes a stub to the server, then it would validate method arguments on that stub?)
> 
> And you're saying capnweb-validate will "rewrite" calls to `newHttpBatchRpcSession<T>()` and similar functions to automatically add validation? How does this rewriting work? The type parameter is obviously needed in order to decide what validation to apply, so runtime monkey-patching can't do the job. I guess you must be using some sort of bundler plugin to do the rewrite?
> 
> I think in many cases this client-side validation is not really needed since the client fully trusts the server. So I'm not sure it's desirable to make it automatic. I would say we should give people an explicit way to add the validation, like:
> 
> ```
> let stub = validateStub<T>(newWebSocketRpcSession<T>(...))`
> ```
> 
> Then this explicit `validateStub` function could also be used on Workers RPC stubs.
> 
> By the way, we probably need a way to coerce a validated stub to another type, so that e.g. if an API returns `RpcStub<any>`, we can coerce it to a specific `RpcStub<T>`. Note the coercion requires _removing_ the existing validation wrapper and then adding a new wrapper. So this may call for another function like `coerceStub<T>` that does this?

yes. today this works through the `capnweb-validate` unplugin/CLI build transform.

the transform uses the TypeScript checker to find calls like `newHttpBatchRpcSession<Api>()`, resolves `Api`, emits a runtime validator, and rewrites the call to an internal helper that calls the real Cap’n Web constructor and wraps the returned stub in a `Proxy`.
that proxy validates outgoing args and resolved return values.
runtime monkey-patching would not work because the type argument is erased.

I agree with the new direction: client-side stub validation should be explicit rather than automatic. I’ll update this PR to remove the automatic Cap’n Web client constructor rewriting. Since server-boundary validation already exists through `@validateRpc()`, the client-side replacement would be an explicit API like `validateStub<T>(stub)`, with a possible `coerceStub<T>()` for replacing an existing wrapper. I can include that here if we agree on the API shape, or split it into a follow-up PR.

teamchong:
@kentonv, I pushed `153d6f6` removing the automatic client constructor rewrite.

so this PR is now only server-boundary validation via `@validateRpc()`. the example reflects that: the Worker is transformed, the React client uses normal Cap’n Web sessions.

for follow-up, I’d make client validation explicit:

```ts
const api = validateStub<Api>(newHttpBatchRpcSession<Api>("/api"));
```

and the same should work for Workers RPC stubs:

```ts
const api = validateStub<Api>(env.SERVICE);
```

I’d have `validateStub<T>()` reuse the old proxy behavior: validate args before calls through the stub, and validate resolved return values on the caller side. for normal client-to-server stubs, the return check is the main value; the arg check matters for callback/capability stubs passed over RPC.

I'd skip `coerceStub<T>()` for now unless we have a concrete `RpcStub<any>` case, and when we do, the unwrap-then-rewrap is the part to get right

kentonv:
> that proxy validates outgoing args and resolved return values.

I don't think we ever need to validate _outgoing_ values. This is redundant since the recipient will need to validate again on its end for security. Can we differentiate outgoing wrappers vs. incoming wrappers, where incoming wrappers validate only arguments and outgoing wrappers validate only return values?

teamchong:
> > that proxy validates outgoing args and resolved return values.
> 
> I don't think we ever need to validate _outgoing_ values. This is redundant since the recipient will need to validate again on its end for security. Can we differentiate outgoing wrappers vs. incoming wrappers, where incoming wrappers validate only arguments and outgoing wrappers validate only return values?

Done. validation is now receive-side only: a wrapper validates what it *receives*, never what it sends, since the receiver re-validates anyway.
- **incoming wrappers** (server target, `@validateRpc()`): validate **arguments** only; returns pass through unvalidated.
- **outgoing wrappers** (client stub, `validateStub<T>()`): validate **resolved returns** only; outgoing args pass through, and the generated client validators omit argument validators entirely.
- capabilities follow the same rule in both directions: a stub you receive and later call through is wrapped as outgoing (validates only its returns); a capability you expose is wrapped as incoming (validates only the args called into it).

client validation is also explicit now. I dropped the automatic Cap'n Web client-constructor rewrite, users opt in with `validateStub<T>(stub)`, which works the same in a worker calling another service

dimitropoulos (approved):

Issue #30, opened 2025-09-23
clarify how bidirectional communication should work


c43721:
Heya, neat library. Almost what I was asking for just a day ago 😆 

I'm having a hard time understanding how to achieve bidirectional communication. It's almost worded like you can just push data back to the client, but I think that's not the right kind of thinking. So I dug deeper I ended up using the functionality that allows you to pass functions by reference- it seems like this is kind of what's intended? 

Where it gets really confusing when you add things like timers into the equation. Without marking the method as `async` and awaiting a promise directly, it doesn't behave as you'd expect (from a "nieve" point of view, that is.)

I have an [example repository ](https://github.com/c43721/test-cap-n-web/tree/2e71235f2630358c2d66b5e5f1715cf86c5ad489) you can use to explore this in more depth

Example client:
```ts
using wsApi = newWebSocketRpcSession<MyWs>("ws://localhost:8000");

const d = (sm: string) => console.log("got something: " + sm)

const wsResult = await wsApi.test(d);

console.log(wsResult);
```

Example websocket "client":
```ts
class WsServer extends RpcTarget {
    async test(fn: (data: string) => void) {
        fn("testing");

        setTimeout(1000, () => {
            fn("hello from timeout");
        });

        return true;
    }
}
```
It'll call the first function, then return `true`. That makes a lot of sense! I would expect this to be a "gotcha" here and not a bug of the protocol (how would you know timers are involved?)

Now replace the `setTimeout` with an import from `node:timers/promises`:

```ts
class WsServer extends RpcTarget {
    async test(fn: (data: string) => void) {
        fn("testing");

        const msg = await setTimeout(1000, "da");
        fn(msg);

        return true;
    }
}
```

We'll get what we want:

```
got something: testing
got something: da
true
```

The ask here is if we can get some examples of bidirectional communication and the "gotchas" those comes with. I can see how functions can be used to achieve the "notification" style API for a client and this is something I'm looking to explore :)

kentonv:
The setTimeout() version actually works fine -- it's just that the callback happens after `test()` has returned. If you change your client code to wait long enough, you'll see the call still happens, asynchronously. Of course, if you ran the some code locally without the RPC system in the middle, you'd get the same behavior.

I agree we could use more examples of callbacks. But there aren't very many "gotchas", it mostly just works exactly like a local callback would.

vilvei:
[...]

PR #172, opened 2026-05-21
docs: add callback retention example to Duplicating stubs

SAY-5:
Refs #30.

In #30, @kentonv (member) said "I agree we could use more examples of callbacks." The follow-up comment showed the typical pitfall: a `setCallback(cbfunc)` method that stores `cbfunc` and invokes it later from a timer throws "Attempted to use RPC stub after it has been disposed" unless the stub is duplicated with `.dup()`.

This adds a short subsection to "Duplicating stubs" in the README with:

- the bidirectional-callback pattern that motivates `.dup()`,
- a 12-line `RpcTarget` example that stores `cbfunc.dup()` and disposes it from `[Symbol.dispose]`,
- a note that the same rule applies in the reverse direction (client retaining a stub returned by the server).

README only. No code changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,6 +379,27 @@ Sometimes you need to pass a stub somewhere where it will be disposed, but also 
 
 Hint: You can call `.dup()` on a property of a stub or promise, in order to create a stub backed by that property. This is particularly useful when you know in advance that the property is going to resolve to a stub: calling `.dup()` on it gives you a stub you can start using immediately, that otherwise behaves exactly the same as the eventual stub would if you awaited it.
 
+#### Holding on to a callback past the call that delivered it
+
+A common bidirectional-calling pattern is for the client to pass a callback to the server, which the server then invokes later (for example from a timer, an event handler, or a subsequent RPC). Because the callback parameter is a stub, and stubs in params are implicitly disposed when the call returns, the server must duplicate the stub with `.dup()` if it wants to invoke the callback after the call completes:
+
+```ts
+class Api extends RpcTarget {
+  async setCallback(cbfunc) {
+    // Without .dup(), `cbfunc` is disposed when setCallback() returns.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -378,6 +378,48 @@ Sometimes you need to pass a stub somewhere where it will be disposed, but also
...
+
+```ts
+import { type RpcStub, RpcTarget } from 'capnweb';
+
+// The callback the client passes in: a stub wrapping a function.
+type CallbackFunc = RpcStub<(msg: string) => void>;
+
+class Api extends RpcTarget {
+  #callbackFunc?: CallbackFunc;
+  #timer?: NodeJS.Timeout;
```
ndisidore: I'm not sure if timers are the best mechanism to showcase this, and it makes this example about 40% longer because of the timer specific bookeeping.

A much simpler example that still gives usage of stub duplications/callbacks would be something like

```ts
import { type RpcStub, RpcTarget } from 'capnweb';

// A callback the client passes in: a stub wrapping a function.
type Listener = RpcStub<(msg: string) => void>;

class Api extends RpcTarget {
  #listener?: Listener;

  // Stubs passed as params are disposed when the call returns, so `.dup()`
  // to keep a reference that outlives registerListener().
  registerListener(listener: Listener) {
    this.#listener?.[Symbol.dispose]();   // release any previous listener
    this.#listener = listener.dup();
  }

  // A *later* call can invoke the retained callback — still valid thanks to .dup().
  notify(msg: string) {
    this.#listener?.(msg);
  }

  // Dispose our duplicate when done so the client-side stub can be freed.
  [Symbol.dispose]() {
    this.#listener?.[Symbol.dispose]();
  }
}
```

And I think this actually does a better job of communicating what is happening here

dimitropoulos: love it! please push and we can merge

Issue #184, opened 2026-05-29
Deserializing untrusted `bigint` values can cause CPU exhaustion


dimitropoulos:
# Description

When deserializing an incoming RPC message, capnweb [passes](https://github.com/cloudflare/capnweb/blob/main/src/serialize.ts#L589-L593) the string portion of a `["bigint", "..."]` wire value directly to the `BigInt()` constructor without any length limit.

`BigInt()` parsing of a decimal string is synchronous and superlinear in the input length, so a peer that sends a very long digit string can tie up the JS event loop for a long time. On Node.js this blocks the process; on Workers the request will hit its CPU limit and be terminated.

This only matters when an application exposes a capnweb endpoint to untrusted peers. Servers that only talk to trusted clients (e.g. local dev / remote-bindings flows) are unaffected in practice.

## Repro

```js
import { newWebSocketRpcResponse } from "capnweb";
// ...accept a WebSocket and hand it to newWebSocketRpcResponse with any target...

// From a client, send a single frame:
ws.send(JSON.stringify(["push", ["bigint", "1".repeat(10_000_000)]]));
// Server blocks for ~hundreds of ms parsing the BigInt before doing anything else.
```

Approximate parse times for `BigInt("1".repeat(n))` on Node 20:

| digits | time |
| --- | --- |
| 1,000,000 | ~40 ms |
| 5,000,000 | ~350 ms |
| 10,000,000 | ~820 ms |

## Suggested fix

Cap the accepted string length (and reject obviously non-numeric input) before calling `BigInt()` in `evaluateImpl`. A few thousand digits is more than enough for any realistic use.

As we do this, we should review that we're properly limiting message sizes overall.

Dr-Emann:
What about "by sending a bajillion 'push' messages, I can OOM a server".

I don't know that capnweb is safe to use to talk to anything untrusted.

PR #185, opened 2026-05-30
Add receiver-side resource limits for untrusted peers (#184)

ndisidore:
## Summary

Deserializing untrusted input could exhaust a peer's resources. Most concretely (#184), a `["bigint", "<digits>"]` wire value was handed straight to `BigInt()` with no length check, and `BigInt()` decimal parsing is superlinear — so a multi-megabyte digit string blocks the event loop. While fixing that, the receive path is also bounded against the related single-message exhaustion vectors the issue calls out ("review that we're properly limiting message sizes overall").

The guards are local, receiver-side decisions that fit capnweb's existing design: no new dependencies, no protocol changes, and reuse of the existing abort path on violation.

## What this changes

- **bigint length cap + numeric validation** — reject digit strings longer than the limit, and reject anything that isn't an optional sign followed by ASCII decimal digits, before calling `BigInt()`.
- **Message nesting depth guard** — `evaluateImpl` now tracks recursion depth and rejects messages nested beyond the limit (mirroring the existing send-side depth cap), so a tiny deeply-nested message can't overflow the stack.
- **Maximum message size** — incoming messages larger than the limit are rejected before `JSON.parse`, bounding the up-front allocation.

## Configuration

Limits are always applied from the exported `DEFAULT_LIMITS`, and can be overridden per session via a new optional `limits` field on `RpcSessionOptions`:

```ts
new RpcSession(transport, target, {
  limits: { maxBigIntDigits: 4096, maxDepth: 32, maxMessageSize: 8 * 1024 * 1024 },
});
```

Defaults:

- `maxBigIntDigits: 16384` — 2^14 digits ≈ ~54,000 bits, far beyond any practical cryptographic integer (an RSA-16384 modulus is ~4,933 decimal digits), so no legitimate value is rejected — yet orders of magnitude below the millions of digits needed for `BigInt()`'s superlinear parse to block meaningfully.
- `maxDepth: 64` — matches the existing send-side depth limit, so capnweb never rejects a message it would itself send.
- `maxMessageSize: 32 MiB` — generous for legitimate batched and base64/blob payloads while still bounding a single allocation.

Because the protocol has no negotiation step and a limit is a purely local receiver-side decision, the defaults are deliberately generous: a sender cannot discover the receiver's limit, so the defaults must never trip legitimate traffic. They are tunable per session for endpoints with unusual needs.

## Behavior on violation

Exceeding any limit throws, which flows through the existing readLoop → `abort()` path: the session sends an `["abort", ...]` frame and tears down. The thrown errors name the limit and its value, so a cooperative peer sees a descriptive reason on the aborted session — riding the abort frame that is already sent, with no protocol change.

## Alternatives considered

- **Hardcoded constants with no override** — simplest, and still covers the standalone `deserialize()` path, but leaves no escape hatch for applications that legitimately exchange very large payloads. Kept the hardcoded defaults as the floor, but made them overridable.
- **Global / static configuration** — rejected: there is no precedent in the codebase (configuration is per-session), it would leak one application's setting across unrelated sessions, and it is awkward inside Workers isolates.
- **Protocol handshake to advertise limits to the peer** — rejected: it adds a round trip (or is structurally impossible for the one-shot HTTP batch transport), undermines promise pipelining by forcing an await before the first call, and adds no security — an untrusted peer simply ignores advertised limits, so the receiver must enforce locally regardless.
- **App-level `getLimits()` RPC method** — purely a cooperative-peer ergonomic that needs no library change (an application can already expose such a method); it is not a substitute for local enforcement, so it is left to applications that want it.

## Deferred (potential follow-ups)

These are larger resource-exhaustion concerns from the issue discussion that need rate/quota or backpressure design rather than a per-message guard, and are intentionally out of scope here:

- Unbounded growth of the imports/exports tables from a flood of `push` / `stream` / `pipe` messages (the "many messages → OOM" case) — needs per-session quotas and backpressure.
- Attacker-chosen import indices causing sparse-array bloat.
- `remap` instruction/capture fan-out producing multiplicative work.
- Streams and blobs buffered fully in memory.

Note: nested call-arguments are deserialized as a fresh payload with a fresh depth budget, so a message that nests deeply *through call arguments* can still recurse beyond `maxDepth`. This fails safe — it is bounded by `maxMessageSize` and produces a clean session abort (a caught `RangeError`), not a process crash (verified) — and belongs to the same follow-up bucket.

## Testing

- New `__tests__/limits.test.ts` (14 tests): just-under / over boundaries for each limit, non-numeric bigint rejection, per-session override enforcement, the standalone `deserialize()` protected by defaults, and backwards-compatibility — sessions constructed with no options, and with empty options, round-trip normal bigints / nesting / calls unchanged.
- Full suite green: **node 156 passed, workerd 145 passed**, no regressions.


ndisidore:

src/serialize.ts
```diff
@@ -576,19 +627,38 @@ export class Evaluator {
...
         case "bigint":
           if (typeof value[1] == "string") {
-            return BigInt(value[1]);
+            let digits = value[1];
+            let maxBigIntDigits = this.importer.getLimits().maxBigIntDigits;
+            if (digits.length > maxBigIntDigits) {
+              throw new TypeError(
+                  `Deserialized bigint exceeds maximum length of ${maxBigIntDigits} digits.`);
+            }
+            // Reject anything BigInt() would parse expensively or unexpectedly before handing it
```
ndisidore: To be discussed:
I think this was a bug/unintentional behavior before. 

> protocol.md:181: "A bigint value, represented as a decimal string."

It *is* technically possible now to handcraft `["bigint","0xff"]` which unconditionally calls `BigInt(value[1])`  meaning this gets parsed as `255n`.
But this means there is a disconnect in the protocol writeup and the actual implementation

.changeset/untrusted-resource-limits.md
```diff
@@ -0,0 +1,5 @@
+---
+"capnweb": patch
```
kentonv: Feels like a "minor" since it adds API features, but then again I never really understood semver.

src/rpc.ts
```diff
@@ -746,6 +767,14 @@ class RpcSessionImpl implements Importer, Exporter {
         }
       }
 
+      // Bound the up-front allocation of a single message before parsing it. This is a local,
+      // receiver-side guard against an untrusted peer sending an enormous message; a throw here
+      // propagates out of readLoop and aborts the session.
+      if (msgText.length > this.limits.maxMessageSize) {
```
kentonv: Feels like it would benefit the underlying transport to know about message size limits so that it could avoid even reading more bytes than the limit into memory.

Could be passed into `receive()` but some transports would probably want to know in the constructor?

Maybe that just means it is up to the app to pass in when they are constructing such a transport.

I wonder if any WebSocket implementations support specifying limits. You'd think they would since otherwise sending enormous WebSocket frames seems like an easy way to exhaust anyone's memory...

src/serialize.ts
```diff
@@ -7,6 +7,47 @@ import { StubHook, RpcPayload, typeForRpc, RpcStub, RpcPromise, LocatedPromise,
...
+  // anyone deploys, is only ~4933 decimal digits). So no legitimate use is rejected. Yet 16384 is
+  // orders of magnitude below the millions of digits needed for BigInt()'s superlinear parse to
+  // cause meaningful blocking, so the DoS vector is closed.
+  maxBigIntDigits: number;
+
+  // Maximum nesting depth of a single deserialized payload tree. Mirrors the send-side limit in
+  // Devaluator.devaluateImpl, so capnweb never rejects a message it would itself produce. Guards
+  // against stack overflow from a deeply-nested payload (e.g. [[[[...]]]]). Note that call
+  // arguments are deserialized as their own payload (with a fresh depth budget), matching how the
+  // send side serializes them; bounding nesting across such boundaries is left to maxMessageSize.
```
kentonv: I actually don't think call arguments should reset the limit. Otherwise that still allows a stack overflow with extremely nested calls.

FWIW matching the sending side is not a priority. The sending side applies a limit only to catch cycles and avoid going into an infinite loop.

src/serialize.ts
```diff
@@ -576,19 +627,38 @@ export class Evaluator {
     return this.evaluate(structuredClone(value));
   }
 
-  private evaluateImpl(value: unknown, parent: object, property: string | number): unknown {
+  private evaluateImpl(
+      value: unknown, parent: object, property: string | number, depth: number): unknown {
+    let maxDepth = this.importer.getLimits().maxDepth;
```
kentonv: There's a lot of separate calls being made to `this.importer.getLimits()`. I think we should store the limits as a private property of the `Evaluator` object -- they aren't expected to change during a single evaluation.

src/serialize.ts
kentonv: Hmm, I'm tempted to allow, and maybe even _require_, the hex format, because it'll obviously be much more efficient to parse. Parsing hex is strictly O(n) so you might even argue we don't need a limit then (though we might as well keep it anyway).

It's nice that switching to generating hex is backwards-compatible...

__tests__/limits.test.ts
```diff
@@ -0,0 +1,236 @@
+// Copyright (c) 2025 Cloudflare, Inc.
```
kentonv: 2026

src/rpc.ts
ndisidore: So "plumb the limit into transports" decomposes into three quite different problems:

- Plumbing: how does the limit even reach a transport? Transports are frequently built by the app and handed to new `RpcSession(transport, …)`, so `options.limits` can't reach a pre-built transport. Only the `newWebSocketRpcSession` / `newHttpBatchRpcResponse` helpers (which construct the transport internally) could thread it. the raw constructor path can't. That's an interface-design decision with back-compat implications
- The socket-based transports (the untrusted-facing ones) can't be enforced by capnweb at all: the platform reads and buffers the frame before capnweb's listener runs. The only real lever is the underlying socket's native limit: Node ws's maxPayload, Bun's maxPayloadLength, or the runtime's own cap (Cloudflare Workers, for instance, already imposes a built-in WebSocket message-size limit). All of those are app-configured when the socket is created  (you mention landing on this as well)
- HTTP batch is the one place capnweb genuinely controls the read, so it's the only transport where the real value add is, and even there it means replacing the clean `.text()` with a manual `ReadableStream `read that counts bytes and aborts past the cap (plus disambiguating batch size vs per-message size, since a batch is newline-joined messages in one body).

My read is this ultimatley boils down to touching the `RpcTransport` interface, all 5 transports, the session-helper entry points, and docs/tests, all while still being unable to protect the most exposed case (a WebSocket server) from capnweb's own code, because that protection physically lives in the socket the app constructs.

I'm going to suggest we leave this to a followup pr

src/serialize.ts
ndisidore: done! :heavy_check_mark:

src/serialize.ts
ndisidore: Done. Changed to now always emit hex, but it can still consume binary (to allow backwards compat)

src/rpc.ts
kentonv: Fair summary. I was thinking something along the lines of passing the limits into the transport constructors. Of course, if the app constructs its own transport, it is its own job to pass limits to that.

But I suppose as you point out the standard WebSocket API doesn't actually give us any ability to configure max incoming message size... we'd have to use non-standard settings if anything... bleh. Maybe leave it up to the app.

.changeset/untrusted-resource-limits.md
```diff
@@ -0,0 +1,7 @@
+---
+"capnweb": minor
+---
+
+Add receiver-side resource limits to guard against untrusted-peer resource exhaustion (#184). Deserialization now caps the length of `bigint` values, accepts both emitted hex strings and legacy decimal strings, bounds message nesting depth across nested call arguments, and rejects oversized incoming messages before parsing. The limits are local, receiver-side decisions with safe defaults (`DEFAULT_LIMITS`), and can be overridden per session via the new `limits` field on `RpcSessionOptions`. Exceeding a limit aborts the session, reusing the existing abort path.
```
kentonv: Hmm, I think the changeset description is really meant to be a one-liner, it shows up in a bullet list in the release notes. I actually don't know how multiple paragraphs will render.

kentonv (approved):

__tests__/limits.test.ts
```diff
@@ -0,0 +1,286 @@
...
+
+// Builds a pipeline call whose sole argument is another pipeline call, repeated `depth` times.
+// This crosses the call-argument RpcPayload boundary at every level.
+function nestedCallArgs(depth: number): string {
+  let s = "1";
+  for (let i = 0; i < depth; i++) {
+    s = `["pipeline",0,["echo"],[${s}]]`;
+  }
+  return s;
+}
```
dimitropoulos, line 24:
```suggestion
function nestedCallArgs(depth: number): string {
  return `["pipeline",0,["echo"],[`.repeat(depth) + "1" + "]]".repeat(depth);
}
```

same as the above - I think we can avoid the loop assignment with this

__tests__/limits.test.ts
```diff
@@ -0,0 +1,286 @@
...
+// Builds the wire string for a value nested inside `depth` escaped-array layers. Each escaped
+// array `[[ inner ]]` adds one level of recursion to the deserializer (evaluateImpl recurses into
+// its single element). The innermost value is the number 1.
+function nestedEscapedArrays(depth: number): string {
+  let s = "1";
+  for (let i = 0; i < depth; i++) {
+    s = `[[${s}]]`;
+  }
+  return s;
+}
```
dimitropoulos, line 18:
```suggestion
function nestedEscapedArrays(depth: number): string {
  return `${"[[".repeat(depth)}1${"]]".repeat(depth)}`;
}
```

maybe I'm misunderstanding the algorithm, but I'm pretty sure it can be this? that way we can avoid needing to do the assignment in a loop. not like it's a big deal or anything

src/rpc.ts
```diff
@@ -746,6 +767,15 @@ class RpcSessionImpl implements Importer, Exporter {
         }
       }
 
+      // Bound a single message before parsing it. At this point the transport has already
+      // buffered the complete string; true pre-read enforcement belongs in the transport/socket.
+      // This backstop still prevents oversized messages from reaching JSON.parse and downstream
+      // deserialization work. A throw here propagates out of readLoop and aborts the session.
+      if (msgText.length > this.limits.maxMessageSize) {
```
dimitropoulos: do we need to be thinking in terms of bytes not characters? so like 

```typescript
// 50 emoji = 100 UTF-16 code units (.length), but 200 bytes in UTF-8
// PASSES the check despite being 2x the "limit" in actual memory
const msg = "😀".repeat(50);
console.log({
  chars: msg.length, // 100
  bytes: new TextEncoder().encode(msg).byteLength, // 200
});

// Conversely, 100 ASCII chars = 100 code units AND 100 bytes
const msg2 = "a".repeat(100);
console.log({
  chars: msg2.length, // 100
  bytes: new TextEncoder().encode(msg2).byteLength, // 100
})
```

__tests__/limits.test.ts
```diff
@@ -0,0 +1,286 @@
...
+    let underLimit = '["bigint","0xff"]';
+    let overLimit = '["bigint","0xfff"]';
+
+    // The cap is applied to the whole bigint string, regardless of radix.
+    await pushFrame(underLimit, { maxBigIntDigits: 4 }).expectNoAbort();
+
+    let aborted = await pushFrame(overLimit, { maxBigIntDigits: 4 }).waitForAbort();
+    expect(aborted).toMatch(/bigint exceeds maximum length/);
+
+    // The same 5-digit bigint is well under the default limit, so it does NOT abort by default.
```
dimitropoulos: but it's not a 5-digit, it's a 3-digit hex number with a `0x` prefix (not part of the number).  In general, I feel like it'd be pretty reasonable to not consider the `0x` in counting digits.

src/serialize.ts
```diff
@@ -576,19 +639,40 @@ export class Evaluator {
...
+            let maxBigIntDigits = this.limits.maxBigIntDigits;
+            if (digits.length > maxBigIntDigits) {
+              throw new TypeError(
+                  `Deserialized bigint exceeds maximum length of ${maxBigIntDigits} characters.`);
+            }
+            if (/^-?[0-9]+$/.test(digits)) {
+              return BigInt(digits);
+            } else if (/^-?0x[0-9a-fA-F]+$/.test(digits)) {
+              let isNegative = digits.startsWith("-");
+              let magnitude = BigInt(isNegative ? digits.slice(1) : digits);
```
dimitropoulos: wow, I didn't know that `BigInt("-0xf")` throws.  huh!

__tests__/limits.test.ts
dimitropoulos: and in reading the section about handling negative makes me wonder if the negative sign would also wrongly be counted as a digit, too.

src/serialize.ts
dimitropoulos: am I understanding correctly?  in the case of a newer server talking to an older client:

- before this PR: `serialize(-255n)` produces `["bigint","-255"]`. Old receivers call `BigInt("-255")` --> works fine.
- after this PR: `serialize(-255n)` produces `["bigint","-0xff"]`. Old receivers call `BigInt("-0xff")` --> throws SyntaxError, aborting the session.

so technically this is a major break?  (well, semver 0.x bump, in our pre-1.0 case).

dimitropoulos (commented):
some questions

__tests__/limits.test.ts
ndisidore: resolved! prefixes stripped

src/rpc.ts
ndisidore: Good spot and a real shout. I suppose the answer is what the pragmatic approach here is given there is some layer of implicit trust between peers

Measured cost of switching .length to byte count (per message, on the hot read path):
Approach | Cost on 1M-cu msg | Notes
-- | -- | --
.length (current) | 0.2 µs | stored property, effectively free
Buffer.byteLength | ~1,240 µs | O(n) scan, no big alloc — but Node/workerd only, not in browsers
TextEncoder().encode().byteLength | ~3,620 µs | O(n) + allocates a full byte-array copy (3 MB here) just to read .byteLength, then discards it
(JSON.parse, for reference) | ~3,530 µs | the work this backstop gates

(buffer is not available in browsers)

Therefore, I think the smart approach is to accept and document this refers to utf-16 code units and live with that for now

dimitropoulos (approved):

PR #186, opened 2026-06-01
Finish RPC transport encoding levels from #144

teamchong:
Builds on @ashkalor’s PR #144 and finishes the remaining review feedback for transport encoding levels.

This keeps the original encoding-level design, with follow-up fixes for the rebase and review comments:

- Type custom transport receive values as `unknown` at the transport boundary.
- Estimate stream message sizes when a custom transport cannot report an encoded size, preserving stream flow-control behavior.
- Preserve async string transport send failures by observing returned promises.
- Preserve default error stack redaction for structured-clone transports.
- Simplify the README custom transport docs and remove the stray character.
- Add a changeset for the new transport encoding-level support.

Testing:

```text
npm run test:types
npm run build
npm test -- __tests__/index.test.ts -t "applies backpressure when custom transport omits stream message size"
npm test -- __tests__/index.test.ts -t "propagates async send failures"
npm test -- __tests__/index.test.ts -t "hides the stack by default with structured clone"
```

Original implementation by @ashkalor in #144.


__tests__/index.test.ts
```diff
@@ -473,6 +474,62 @@ class TestTransport implements RpcTransport {
   }
 }
 
+class ObjectTestTransport implements RpcTransportWithCustomEncoding {
+  constructor(
+      private partner?: ObjectTestTransport,
+      readonly encodingLevel: "json" | "jsonWithBytes" | "structuredClone" = "json") {
```
dimitropoulos: seems like it'd be nice to have
```ts
type CustomEncodingLevel = "json" | "jsonWithBytes" | "structuredClone";
```

```suggestion
      readonly encodingLevel: CustomEncodingLevel = "json") {
```

README.md
```diff
@@ -776,3 +776,10 @@ let stub: RemoteMainInterface = session.getRemoteMain();
 ```
 
+
+By default, `send()` accepts a string, and `receive()` returns a string, with Cap'n Web handling the encoding all the way to and from strings. However, transports that want more control over the serialization can declare the property `encodingLevel` to control just how much encoding Cap'n Web does before passing off the message:
```
dimitropoulos:
```suggestion
By default, `send()` accepts a string, and `receive()` returns a string, with Cap'n Web handling the encoding all the way to and from strings. However, transports that want more control over the serialization can declare the property `encodingLevel` to control how much encoding Cap'n Web does before passing off the message:
```

README.md
```diff
@@ -776,3 +776,10 @@ let stub: RemoteMainInterface = session.getRemoteMain();
...
+
+* `"string"`: The default. Messages are strings.
```
dimitropoulos:
```suggestion
* `"string"` (default): Full JSON round-trip. The transport deals in strings only. Cap'n Web handles all encoding/decoding. This is what HTTP batch and WebSocket transports use.
```

README.md
```diff
@@ -776,3 +776,10 @@ let stub: RemoteMainInterface = session.getRemoteMain();
...
+* `"json"`: Messages are JSON-compatible objects. The transport is responsible for serializing/deserializing.
```
dimitropoulos:
```suggestion
* `"json"`: The transport works with JavaScript object trees, but they must be JSON-compatible. Cap'n Web still encodes special types, but skips the final JSON.stringify. The transport is responsible for serialization (e.g. to CBOR, MessagePack).
```

README.md
```diff
@@ -776,3 +776,10 @@ let stub: RemoteMainInterface = session.getRemoteMain();
...
+* `"jsonWithBytes"`: Like "json" except that byte arrays are left as `Uint8Array` instead of base64-encoded. Handy for use with serializations like CBOR or MessagePack that support this efficiently.
```
dimitropoulos:
```suggestion
* `"jsonWithBytes"`: Like "json" except that byte arrays are left as `Uint8Array` instead of base64-encoded, avoiding the ~33% base64 size overhead and the encode/decode CPU cost. Handy for use with serializations like CBOR or MessagePack that support this efficiently.
```

README.md
```diff
@@ -776,3 +776,10 @@ let stub: RemoteMainInterface = session.getRemoteMain();
...
+* `"structuredClone"`: Messages are structured-clonable objects. Cap'n Web will only implement special handling of RPC stubs. This is useful when the transport is a `MessagePort` or similar.
```
dimitropoulos:
```suggestion
* `"structuredClone"`: Messages are [structured-clonable](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#supported_types) objects. Cap'n Web will only implement special handling of RPC stubs. This is useful when the transport is a `MessagePort` or similar.
```

dimitropoulos (commented):
code looks good, but personally I think it could be confusing that "string" means "json string" and "json" means "objects that could be serialized to a json string - but not actually a json string".  I think of json as a serialization format, and when I see "json" as an option I'd wrongly think it means what "string" actually means.

my suggestion would be:

`string` -> `json` (possibly, more clear, `jsonStringified` or `jsonString`)
`json` -> `jsonCompatible`
`jsonWithBytes` -> `jsonCompatibleWithBytes`
`structuredClone` -> good! but maybe there's an argument to be made for `structuredClonable`

README.md
teamchong:
```
* `"structuredClone"`: Messages are structured-clonable values. Cap'n Web passes through native structured-clone types where possible, while still handling RPC-specific values such as stubs. This is useful when the transport is a `MessagePort` or similar.
```

I tweaked this instead of applying verbatim because “only RPC stubs” is too narrow. Even at `structuredClone`, Cap’n Web still does protocol-specific handling beyond stubs, so this wording is meant to be more accurate while keeping your structured-clonable framing.

teamchong:
bonk's working now, the run only failed on a git credential thing after the review, I included the fixes in this PR

src/serialize.ts
```diff
@@ -7,6 +7,26 @@ import { StubHook, RpcPayload, typeForRpc, RpcStub, RpcPromise, LocatedPromise,
...
+ * @example
+ * ```ts
+ * // What happens to Uint8Array([1, 2, 3]) at each level:
+ * "string"          → '["bytes","AQID"]'           // JSON string with base64
+ * "json"            → ["bytes", "AQID"]            // JS array with base64
+ * "jsonWithBytes"   → ["bytes", Uint8Array]        // JS array with raw bytes
+ * "structuredClone" → ["bytes", Uint8Array]        // + Date, BigInt stay native
+ * ```
+ */
+export type EncodingLevel = "string" | "json" | "jsonWithBytes" | "structuredClone";
```
dimitropoulos: we talked in our meeting, and let's rename some of these:

json -> jsonCompatible
jsonWithBytes -> jsonCompatibleWithBytes
structuredClone -> structuredClonable

dimitropoulos (changes requested):

.changeset/quiet-dragons-hunt.md
```diff
@@ -0,0 +1,7 @@
+---
+"capnweb": minor
```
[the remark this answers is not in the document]
teamchong: Keep as minor

src/serialize.ts
teamchong: updated

dimitropoulos (approved):

PR #189, opened 2026-06-03
Add WebSocket serialization support

jonastemplestein:
## What changed

Adds support for passing live WebSockets through Cap'n Web RPC messages by exporting the socket as a capability and reconstructing a bridged WebSocket-like object on the receiving side.

This also allows `Response` values carrying a `.webSocket` property to cross RPC boundaries. The presence of `.webSocket` is the upgrade signal; the code does not depend on status `101`.

## Why

This enables a Cap'n Web `fetch(request)` target to return a WebSocket-bearing `Response`, so callers can run WebSocket traffic, including another Cap'n Web WebSocket session, through an existing Cap'n Web tunnel.

## Scope notes

- Adds the `websocket` wire expression plus the 4-element `response` form to `protocol.md`.
- Adds a short README note describing WebSocket pass-by-reference semantics, bridge shape, and disposal expectations.
- Includes one focused Node test file covering direct WebSocket serialization, binary frames, close code/reason propagation, `Response.webSocket`, and Cap'n-Web-over-bridged-WebSocket.
- Kept the public quick-tunnel proof out of the default test suite because it depends on external network, `cloudflared`, and DNS propagation.
- The only general test infrastructure change binds the existing test server to `127.0.0.1`; this makes the existing WebKit browser HTTP/WebSocket tests reachable in the local full matrix.

## Review follow-ups applied

- Removed scratch/public-network test files from the PR.
- Avoided lockfile churn and unrelated test skips.
- Fixed bridge lifetime so `close()` preserves the retained receiver until the remote close event arrives.
- Made bridge startup one-shot to avoid leaking a replaced retained receiver.
- Preserved `Response.webSocket` on local `Response` copies.
- Normalized binary frame payloads and accepted ArrayBuffer / typed-array sends.
- Strip only invalid 1xx response status/statusText when serializing WebSocket-bearing responses.

## Validation

- `npx vitest run --project node __tests__/websocket-serialization.test.ts` - 2 passed
- `npm run test:types`
- `npm run build`
- `npm run test:bun` - 16 passed
- `npm test` - 817 passed
- Manual live public tunnel proof: `CAPNWEB_PUBLIC_TUNNEL=1 npx vitest run --project node __tests__/websocket-public-tunnel.test.ts` passed locally before removing the external-network test from the PR. It opened a real `trycloudflare.com` tunnel, connected via public WSS, received a WebSocket-bearing `Response`, and ran Cap'n Web over that bridged socket.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,6 +205,14 @@ The following types can be passed over RPC (in arguments or return values), and 
 * `ReadableStream` and `WritableStream`, with automatic flow control.
 * `Headers`, `Request`, and `Response` from the Fetch API.
 
+Live `WebSocket` objects, including a Cloudflare Workers `Response` with a `.webSocket`
+property, can also be passed over RPC. Unlike the types above, a WebSocket is passed by
+reference: the receiver gets a WebSocket-like bridge connected to the original socket. The bridge
+supports `send()`, `close()`, `accept()`, `readyState`, and `message` / `close` / `error` event
+listeners, but it is not a full platform WebSocket object. Close or dispose it when done so the
+remote socket can be released. WebSocket tunneling carries frames over the RPC session and does
+not add `ReadableStream`-style flow control.
+
 The following types are not supported as of this writing, but may be added in the future:
 * `Map` and `Set`
 * `ArrayBuffer` and typed arrays other than `Uint8Array`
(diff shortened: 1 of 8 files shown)


jonastemplestein:
Superseded by the fork-local PR: https://github.com/iterate/capnweb/pull/1

PR #190, opened 2026-06-05
Fix five correctness and robustness issues in serialization and transport handling

taylorodell:
This PR fixes five small, independent correctness/robustness bugs in serialization and transport handling, and adds a focused unit test for each.

### 1. Error type names should not resolve to `Object.prototype` members
`src/serialize.ts` (the `"error"` case): the wire-supplied error type name (`value[1]`) is looked up in `ERROR_TYPES`, which was a plain object literal inheriting `Object.prototype`. A name like `"constructor"` resolved to `Object.prototype.constructor` (i.e. `Object`) — which is truthy, so the `|| Error` fallback was skipped — and `new Object(message)` then produced a `String` wrapper rather than an `Error`, defeating downstream `instanceof Error` checks while still carrying the sender's own-property bag. A name like `"toString"` resolved to a non-constructor and threw, aborting the session. `ERROR_TYPES` now has a null prototype, so unknown or inherited names resolve to `undefined` and correctly fall back to `Error`.

### 2. Error deserialization should filter `Object.prototype` keys
`src/serialize.ts` (the `"error"` case): the plain-object deserializer already skips keys that shadow `Object.prototype` members (and `toJSON`), but the error path only skipped `name`/`message`/`stack`. As a result, keys like `__proto__`, `toString`, `valueOf`, `hasOwnProperty`, or `toJSON` in an error's own-property bag were assigned as own properties on the deserialized error (and `__proto__` in particular went through the prototype setter). This makes the error path consistent with the plain-object path: such keys are skipped, while their inner values are still evaluated so any stubs are released.

### 3. Guard against double-resolving an import
`src/rpc.ts`, `ImportTableEntry.resolve()`: if `resolve()` is ever called a second time, `this.resolution` was overwritten without disposing the previous hook, leaking it. Added a guard at the top that disposes the incoming hook and returns when a resolution already exists.

### 4. `abort` handler passes the wrong value
`src/rpc.ts`, the `"abort"` case in `readLoop()`: it passed the `RpcPayload` wrapper to `this.abort()` instead of the unwrapped reason. The adjacent `"reject"` handler already uses `payload.value`; this change makes `abort` consistent, so error handlers receive the actual reason rather than the wrapper.

### 5. Truncate WebSocket close reason to 123 bytes
`src/websocket.ts`, `WebSocketTransport.abort()`: `WebSocket.close(code, reason)` requires the reason to be ≤ 123 UTF-8 bytes; a longer reason makes `close()` throw and escape the abort path. The reason is now truncated on a UTF-8 character boundary (using a streaming `TextDecoder` so no partial code unit is emitted) before calling `close()`.

### Testing
Added one unit test per fix in `__tests__/index.test.ts` (existing test style/harness). All existing and new tests pass via `vitest run` (`node` and `workerd` projects locally; the `@vitest/browser` projects run on CI). A patch changeset is included.


taylorodell:

src/websocket.ts
```diff
@@ -117,6 +117,14 @@ class WebSocketTransport implements RpcTransport {
     } else {
       message = `${reason}`;
     }
+    // WebSocket close reasons are limited to 123 UTF-8 bytes; a longer string makes
+    // close() throw. Truncate on a character boundary to stay within the limit.
+    let reasonBytes = new TextEncoder().encode(message);
+    if (reasonBytes.length > 123) {
```
ndisidore, line 130: Do you mind explaining this 123 number? Where did this come from and how did you arrive at this number?

kentonv: Looks like it's in the docs: https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/close#reason

TIL

kentonv (approved):

kentonv:
So sorry for the long delay here, the month of June was extremely busy for me and I basically stopped reading my email. Should have looked at this much sooner, as these are important bugs to fix. Thank you!

src/serialize.ts
```diff
@@ -777,6 +785,13 @@ export class Evaluator {
               let propsObj = <Record<string, unknown>>props;
               for (let key of Object.keys(propsObj)) {
                 if (key === "name" || key === "message" || key === "stack") continue;
+                if (key in Object.prototype || key === "toJSON") {
+                  // Consistent with the plain-object deserializer below: don't allow error
+                  // properties to override Object.prototype members (e.g. __proto__, toString,
+                  // valueOf) or toJSON. Still evaluate the inner value so any stubs are released.
+                  this.evaluateImpl(propsObj[key], result, key);
```
dimitropoulos:
```suggestion
                  this.evaluateImpl(propsObj[key], result, key, depth + 1);
```

PR #194, opened 2026-06-10
fix(validate): support decorated classes extending decorated classes

teamchong:
## Problem

A downstream repo hit:

```
Uncaught TypeError: capnweb-validate: Class1.field1 is not in the generated validator
```

and calls to subclass-only methods failed with the same misleading error.

**Root cause:** `@validateRpc()` returned a `new Proxy(...)` from the decorated constructor. When a decorated class extends another decorated class (`Class2 extends Class1`), `super()` returns the base class's validator proxy, so every subclass-only method (`method2`, `method3`, ...) is checked against the base validator and refused. The generated validator itself is correct; only the runtime wrapping was wrong.

The constructor-returns-Proxy approach had a second cost: instances were Proxies rather than real branded `RpcTarget`s, a serialization hazard over native workerd RPC.

## Fix

- `__validateRpcClass` now wraps the declared methods in place on the class's prototype instead of returning a Proxy from the constructor.
- Wrapping is idempotent (a `WRAPPED_METHOD` symbol marks wrapped functions), so decorated-extends-decorated and double-decoration compose: each class's methods are validated against its own validator, and inheritance behaves like ordinary JS.
- Instances stay real branded `RpcTarget`s with intact `instanceof` identity and `#`-field privacy.
- Incoming argument stubs (callbacks declared with `v.stubOf(...)`) now pass through as native stubs instead of being wrapped in validation proxies. The wrapper proxy could not be forwarded over native workerd RPC (for example, handing a received callback to a Durable Object). The trade-off: return values from those callbacks are no longer validated by default. This follows the review direction on #169: incoming wrappers validate only arguments, and stub validation is explicit opt-in via `validateStub<T>(stub)` rather than automatic.
- Refusal errors now distinguish instance-property access from methods genuinely missing from the RPC interface, and suggest rebuilding if the validator may be stale.

## Verification

- Full `capnweb-validate` suite: 179/179 pass. New regression file `decorated-class-runtime.test.ts` covers subclass-only methods, inherited methods, double-decoration, prototype/identity preservation, and refusal semantics.
- workerd suite: 146/146 pass, including a new integration test that receives a native `RpcStub` callback through a validated service and forwards it to a Durable Object, both as a top-level argument and nested inside an object. The test fails against the previous wrapping behavior.
- Return direction over native workerd RPC, verified with the same wrangler repro: a decorated method whose declared return type resolves to a stub-shaped validator (a class type) returns the instance across the boundary correctly, and declared methods on the returned stub still validate because the prototype methods are wrapped in place. Interface-typed returns emit object shapes and pass through unchanged. Over capnweb sessions the returned stub keeps the validation wrapper, preserving nested capability validation per the #169 direction (receiver validates, sender wraps nested capabilities).
- Minimal repro (wrangler dev over native Workers RPC + a pure capnweb session), before vs after:

| Probe | Before | After |
|---|---|---|
| Subclass-only method on decorated-extends-decorated | fails: `method2 is not in the generated validator` | passes |
| Top-level return of decorated instance | Proxy | real branded `RpcTarget` |
| Forward a received callback stub to a Durable Object | fails in serialization | passes |
| Instance-property probe (`field1`) | refused | refused (clearer message) |
| Inherited method via subclass | passes | passes |
| Decorated method returns a decorated instance over native RPC | Proxy instance, serialization hazard | passes, declared methods still validated |

Note: the downstream `Class1.field1` console error is independently legitimate. `protected` fields are runtime-public properties and should be `#`-private in the downstream code; this PR makes the refusal message say so explicitly.

### DIFF
--- a/packages/capnweb-validate/__tests__/decorated-class-runtime.test.ts
+++ b/packages/capnweb-validate/__tests__/decorated-class-runtime.test.ts
@@ -1,0 +1,237 @@
+// Copyright (c) 2026 Cloudflare, Inc.
+// Licensed under the MIT license found in the LICENSE.txt file or at:
+//     https://opensource.org/license/mit
+
+// Runtime behavior of `__validateRpcClass` (the runtime half of
+// `@validateRpc()`). The decorator must wrap declared methods in place on the
+// class's prototype rather than Proxy-wrapping instances: native (workerd)
+// RPC refuses Proxy targets, and a decorated subclass of a decorated base
+// must resolve subclass-only methods against the subclass validator.
+// Regression tests for the GitHubPullRequestImpl-extends-GitHubIssueImpl
+// failure ("X is not in the generated validator").
+
+import { describe, expect, it } from "vitest";
+import {
+  __validateRpcClass,
+  isWrappedMethod,
+  v,
+  wrapServerTarget,
+  type ServiceValidator,
+  type Validator,
+} from "../src/internal/core.js";
+
+function countingString(counter: { calls: number }): Validator {
+  return ((value: unknown, path) => {
+    counter.calls++;
+    v.string(value, path);
+  }) as Validator;
+}
+
+describe("__validateRpcClass prototype wrapping", () => {
+  it("constructs real instances: identity, prototype chain, and # fields intact", async () => {
+    let captured: unknown;
+    class Svc {
+      #greeting = "pong";
+      constructor() {
+        captured = this;
+      }
+      async ping(): Promise<string> {
+        return this.#greeting;
+      }
+    }
+    const validator: ServiceValidator = {
+      serviceName: "Svc",
+      methods: { ping: { args: [], returns: v.string } },
+    };
+    const Decorated = __validateRpcClass(validator)(Svc);
+
+    let instance = new Decorated();
+    // The old implementation returned a Proxy from the constructor, so the
+    // value you got from `new` was not the `this` the constructor saw, and
+    // the prototype chain gained a synthetic subclass.
+    expect(instance).toBe(captured);
+    expect(Object.getPrototypeOf(instance)).toBe(Decorated.prototype);
+    expect(instance instanceof Svc).toBe(true);
+    // `#` field access works because methods run on the real instance.
+    await expect(instance.ping()).resolves.toBe("pong");
+    // The prototype method is the validated wrapper.
+    expect(isWrappedMethod(Decorated.prototype.ping)).toBe(true);
+  });
+
+  it("decorated subclass of a decorated base resolves subclass-only methods", async () => {
+    class IssueImpl {
+      #title = "hello title";
+      async getTitle(): Promise<string> {
+        return this.#title;
+      }
+    }
+    const issueValidator: ServiceValidator = {
+      serviceName: "IssueImpl",
+      methods: { getTitle: { args: [], returns: v.string } },
+    };
+    const VIssue = __validateRpcClass(issueValidator)(IssueImpl);
+
+    class PullImpl extends VIssue {
+      async readDiff(): Promise<string> {
+        return "diff content";
+      }
+    }
+    const pullValidator: ServiceValidator = {
+      serviceName: "PullImpl",
+      methods: {
+        // Generated subclass validators repeat inherited methods.
+        getTitle: { args: [], returns: v.string },
+        readDiff: { args: [], returns: v.string },
+      },
+    };
+    const VPull = __validateRpcClass(pullValidator)(PullImpl);
+
+    let pull = new VPull();
+    // Regression: previously rejected with
+    // "PullImpl.readDiff is not in the generated validator" because the base
+    // class's instance Proxy answered for the subclass.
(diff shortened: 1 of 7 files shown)


kentonv:
Let's plan on @dimitropoulos reviewing changes to this package, I probably don't have bandwidth to really learn the code myself.

packages/capnweb-validate/__tests__/decorated-class-runtime.test.ts
```diff
@@ -0,0 +1,237 @@
...
+      methods: { title: { args: [], returns: v.string, isGetter: true } },
+    })(WithGetter);
+
+    let instance = new Decorated();
+    expect(instance.title).toBe("hello");
+    let desc = Object.getOwnPropertyDescriptor(Decorated.prototype, "title");
+    expect(isWrappedMethod(desc?.get)).toBe(true);
+  });
+
+  it("leaves undeclared prototype methods callable (exposure is the RPC layer's model)", async () => {
```
dimitropoulos, line 190: ok - so they _weren't_ callable before, right?

packages/capnweb-validate/src/internal/core.ts
```diff
@@ -1259,11 +1289,111 @@ export function __validateRpcClass<T extends new (...args: any[]) => object>(
...
+    // Unchecked members (overloads, platform hooks) stay raw by design.
+    if (isUncheckedMethod(methodSpec)) continue;
+    // Find the implementation wherever it lives on the chain. If it's
+    // inherited from an undecorated base we shadow it with a wrapper on this
+    // prototype; if it's inherited from a decorated base it's already
+    // wrapped and the base's validator governs it.
+    let desc = exposedDescriptor(proto, prop);
+    if (!desc) continue; // Declared but implemented as an instance member (or
+    // absent): there's nothing on the prototype to wrap, and the RPC layers
+    // refuse instance properties themselves.
```
dimitropoulos: don't bother changing, but just thought this was one of the strangest comment structure I've seen in a long time.  LLMs are weird, haha.  

maybe we should get a formatter on this repo some time

packages/capnweb-validate/src/internal/core.ts
```diff
@@ -1259,11 +1289,111 @@ export function __validateRpcClass<T extends new (...args: any[]) => object>(
...
+    return value;
   };
 }
+
+/**
+ * Marks prototype methods that `__validateRpcClass` has already wrapped, so
+ * decorating a subclass of a decorated base (or decorating twice) never
+ * double-validates, and session wrappers can pass such methods through.
+ */
+const WRAPPED_METHOD = Symbol.for("capnweb-validate.wrappedMethod");
```
dimitropoulos, line 1314: do we care that any external thing could also check for this (since it's in the global symbol registry)?  could that be some kind of security problem?  I don't think it is - but just thought I'd call it out.

packages/capnweb-validate/src/internal/core.ts
```diff
@@ -841,13 +856,15 @@ function wrapResolvedValue(
         key,
         shape.key,
         [...path, i, "key"],
-        side
+        side,
+        wrapStubs
       );
       let wrappedValue = wrapResolvedValue(
```
dimitropoulos, line 862: this is getting to be a lot of positionals.  is now the time to object-wrap?  up to you.  I guess it's fine as is too - not required.  also, probably faster in the grand scheme of things, I'd imagine.

packages/capnweb-validate/src/internal/core.ts
```diff
@@ -606,7 +606,11 @@ function exposedDescriptor(
 
 function missingMethod(serviceName: string, prop: string): never {
   throw newValidationTypeError(
-    `capnweb-validate: ${serviceName}.${prop} is not in the generated validator`
+    `capnweb-validate: refused ${serviceName}.${prop}: it is not declared on ` +
```
dimitropoulos, line 609: not a big deal at all but "template + template + template" feels sorta silly - but maybe this is another coin in the "let's add a formatter" bucket

packages/capnweb-validate/__tests__/decorated-class-runtime.test.ts
teamchong: right, externally refused, internally allowed under the old Proxy.
now local calls are uniformly allowed, and refusal of undeclared members lives solely at the RPC boundary, which is the boundary that actually matters for exposure

dimitropoulos (approved):
some very very minor things I noticed - none blocking

packages/capnweb-validate/src/internal/core.ts
teamchong: fixed

packages/capnweb-validate/src/internal/core.ts
teamchong: not a security boundary, symbols don't cross the wire. Symbol.for so duplicate bundled copies recognize each other's wrappers. added a note

packages/capnweb-validate/src/internal/core.ts
teamchong: kept positional, hot recursive path

PR #195, opened 2026-06-11
fix(batch): add missing early return in nodeHttpBatchRpcResponse on non-POST

aleister1102:
## Summary

`nodeHttpBatchRpcResponse` wrote a 405 status for non-POST requests via `response.writeHead(405, ...)` but was missing `response.end()` and `return`. Execution fell through into the full RPC pipeline unconditionally, then crashed with `ERR_HTTP_HEADERS_SENT` when the function later tried to write the 200 response — leaving the TCP connection open until timeout.

The Fetch API counterpart (`newHttpBatchRpcResponse`) was unaffected; it already returned a 405 `Response` immediately.

## Changes

- `src/batch.ts`: add `response.end()` and `return` after the 405 `writeHead` call in `nodeHttpBatchRpcResponse`

## Test plan

- [ ] Verify a non-POST request to a `nodeHttpBatchRpcResponse` endpoint now receives a 405 with a closed connection and no RPC handlers are invoked
- [ ] Verify a valid POST request continues to work normally

### DIFF
--- a/.changeset/non-post-early-return.md
+++ b/.changeset/non-post-early-return.md
@@ -1,0 +1,6 @@
+---
+"capnweb": patch
+---
+
+Fix nodeHttpBatchRpcResponse leaving the connection open and crashing with
+ERR_HTTP_HEADERS_SENT on non-POST requests. It now returns 405 immediately.
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -2069,6 +2069,12 @@ describe("HTTP requests", () => {
     expect(await Promise.all([promise1, promise2, promise3]))
         .toStrictEqual([36, 5, 9]);
   });
+
+  it("rejects non-POST requests with 405", async () => {
+    let response = await fetch(`http://${inject("testServerHost")}`, { method: "GET" });
+    expect(response.status).toBe(405);
+    await response.text();
+  });
 });
 
 describe("WebSockets", () => {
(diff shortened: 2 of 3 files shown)


aleister1102:

aleister1102:
Hi @teamchong @kentonv, could you take a look when you get a chance? This is the fix for the missing early return discussed in the HackerOne report. The change is minimal - two lines in `src/batch.ts`. Thanks!

teamchong:
verified the fix locally. on main a GET to this endpoint never gets a response, the socket just stays open until timeout, and there's an unhandled ERR_HTTP_HEADERS_SENT rejection later when the pipeline runs anyway.

with your branch it returns the 405 immediately. So the change is correct.

two things to add before this can merge (snippets below are just a starting point, adjust as you see fit):

a changeset (the release automation needs it):

`.changeset/non-post-early-return.md`
```md
---
"capnweb": patch
---

Fix nodeHttpBatchRpcResponse leaving the connection open and crashing with
ERR_HTTP_HEADERS_SENT on non-POST requests. It now returns 405 immediately.
```

and a regression test, since this path had no coverage. In `__tests__/index.test.ts`, inside the existing `describe("HTTP requests")` block:

```ts
  it("rejects non-POST requests with 405", async () => {
    let response = await fetch(`http://${inject("testServerHost")}`, { method: "GET" });
    expect(response.status).toBe(405);
    await response.text();
  });
```

this hangs on main and passes with your fix, so it actually catches the regression.

Thanks!

aleister1102:
Done - thanks for verifying the fix locally, @teamchong.

Added both items:
• Changeset: `.changeset/non-post-early-return.md`
• Regression test: rejects non-POST requests with 405 in the existing describe ("HTTP requests") block

Confirmed locally that the new test passes on this branch. Should be ready for another look.

kentonv (approved):

kentonv:
Hmm looks good to me but the new test seems to be failing.

PR #197, opened 2026-06-14
Skip validation for fetch and connect lifecycle methods on WorkerEntrypoint and DurableObject.

teamchong:
`@validateRpc` and the server markers (`newWorkersRpcResponse` etc.) treat `fetch` and `connect` as platform lifecycle entrypoints on `WorkerEntrypoint` and `DurableObject` targets: the Workers runtime invokes them directly rather than over RPC, so they pass through unvalidated instead of being validated as RPC methods.

The platform-passthrough filtering used by the decorator path is now factored into a shared `applyPlatformPassthrough` and also applied on the server-marker call-site path, which previously did no platform filtering. `fetch` is a declared, overridable hook on both base types; `connect` is not declared on either, so it is matched by name within a source-verified Workers-base gate.

Tests cover the marker path for both base types (including a subclass chain), a package-shaped `@cloudflare/workers-types` fixture that exercises the real symbol-origin checks, and a negative case proving a non-Workers `RpcTarget.connect` stays validated.

### DIFF
--- a/packages/capnweb-validate/src/transform/type-introspector.ts
+++ b/packages/capnweb-validate/src/transform/type-introspector.ts
@@ -1245,12 +1245,25 @@ function isRpcReadableProperty(
   );
 }
 
+// Lifecycle entrypoints the Workers runtime invokes directly, not over RPC.
+// Matched by name because `connect` is not declared on either base type.
+const WORKERS_LIFECYCLE_METHOD_NAME_FALLBACKS = ["fetch", "connect"] as const;
+
+function isWorkersLifecycleBaseSymbol(sym: ts.Symbol): boolean {
+  let name = sym.getName();
+  return (
+    (name === "WorkerEntrypoint" || name === "DurableObject") &&
+    isRpcRuntimeSymbol(sym)
+  );
+}
+
 // Platform-inherited method names of `type` (e.g. WorkerEntrypoint `fetch`).
 export function collectPlatformMethodNames(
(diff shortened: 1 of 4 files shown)


packages/capnweb-validate/src/transform/type-introspector.ts
```diff
@@ -1245,12 +1245,25 @@ function isRpcReadableProperty(
   );
 }
 
+// Lifecycle entrypoints the Workers runtime invokes directly, not over RPC.
+// Matched by name because `connect` is not declared on either base type.
+const WORKERS_LIFECYCLE_METHOD_NAME_FALLBACKS = ["fetch", "connect"] as const;
```
dimitropoulos, line 1250: connect is now treated as a passthrough solely by name for any target that extends WorkerEntrypoint or DurableObject, because it is not declared on the Workers base types. That means an ordinary user RPC method named connect on one of those classes, even one included via an explicit surface, would silently be removed from validation and exposed as passthrough.

If connect is intended to be fully reserved for Workers-backed targets, could we document that or emit a clearer diagnostic? Otherwise, this probably needs a tighter gate than name alone, ideally based on a recognizable platform signature/source once the types expose it.

dimitropoulos: nevermind! :)

dimitropoulos (approved):

PR #200, opened 2026-06-24
fix(tests): bind test server to loopback so Firefox/WebKit can connect

SamJB123:
The global test server called `httpServer.listen(0)` with no host, so on dual-stack hosts Node binds the IPv6 wildcard `::`. That address is then provided to clients verbatim as `testServerHost`, so tests connect to `http://[::]:PORT` / `ws://[::]:PORT`.

Chromium tolerates connecting to the wildcard `[::]`, but Firefox and WebKit reject it, causing the `browsers-without-using` project to fail deterministically on those engines:

  - HTTP requests > can perform a batch HTTP request  (NetworkError / Load failed)
  - WebSockets > can open a WebSocket connection      (WebSocket connection failed)

Bind explicitly to `127.0.0.1` so every engine gets a concrete loopback target. Passing a host makes `listen()` take the async DNS-resolution path, so we await the "listening" event before reading `address()` (with no host, binding was synchronous).

Verified on macOS: Firefox and WebKit go from failing those two tests to passing; node and workerd are unaffected.

### DIFF
--- a/__tests__/test-server.ts
+++ b/__tests__/test-server.ts
@@ -50,8 +50,14 @@ export async function setup(project: TestProject) {
     newWebSocketRpcSession(ws as any, new TestTarget());
   })
 
-  // Listen on an ephemeral port for testing purposes.
-  httpServer.listen(0);
+  // Listen on an ephemeral port for testing purposes. Bind explicitly to the IPv4 loopback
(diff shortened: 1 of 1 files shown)


SamJB123:

kentonv (approved):
I don't quite understand why this doesn't fail for me, but the change looks harmless.

kentonv:
Thanks!

PR #201, opened 2026-06-25
feat: Support ArrayBuffer and Buffer Views serialization

ttmx:
Added serialization support for ArrayBuffer and Buffer views, like Int8Array, Uint8ClampedArray, etc.
- Serialized as [bytes, b64, Int8Array/Uint8ClampedArray...], keeping [bytes, b64] for backwards compat with Uint8Array
- Fixes https://developers.cloudflare.com/email-service/local-development/sending/#binary-attachments and previous issues with AI Search

### DIFF
--- a/protocol.md
+++ b/protocol.md
@@ -170,9 +170,12 @@ The values Infinity, -Infinity, and NaN.
 
 `["bytes", base64]`, `["bytes", base64, type]`
 
-A `Uint8Array`, `ArrayBuffer`, `DataView`, or typed array, represented as a base64-encoded string.
-The optional third element preserves the byte container type across the wire. If it is
-omitted, the receiver should deserialize bytes as its default `Uint8Array` for backwards compatibility.
+A byte container, represented as a base64-encoded string. If `type` is omitted, the receiver
+should deserialize bytes as its default `Uint8Array` for backwards compatibility. Otherwise,
+`type` preserves the byte container type across the wire. The supported `type` values are
+`ArrayBuffer`, `DataView`, `Int8Array`, `Uint8ClampedArray`,
+`Int16Array`, `Uint16Array`, `Int32Array`, `Uint32Array`, `BigInt64Array`, `BigUint64Array`,
+`Float32Array`, and `Float64Array`.
 
 `["blob", type, readableExpression]`
 
--- a/protocol.md
+++ b/protocol.md
@@ -175,7 +175,8 @@ should deserialize bytes as its default `Uint8Array` for backwards compatibility
 `type` preserves the byte container type across the wire. The supported `type` values are
 `ArrayBuffer`, `DataView`, `Int8Array`, `Uint8ClampedArray`,
 `Int16Array`, `Uint16Array`, `Int32Array`, `Uint32Array`, `BigInt64Array`, `BigUint64Array`,
-`Float32Array`, and `Float64Array`.
+`Float32Array`, and `Float64Array`. Multi-byte typed array elements are encoded in little-endian
+byte order.
 
 `["blob", type, readableExpression]`
 
--- a/protocol.md
+++ b/protocol.md
@@ -173,7 +173,7 @@ The values Infinity, -Infinity, and NaN.
 A byte container, represented as a base64-encoded string. If `type` is omitted, the receiver
 should deserialize bytes as its default `Uint8Array` for backwards compatibility. Otherwise,
 `type` preserves the byte container type across the wire. The supported `type` values are
-`ArrayBuffer`, `DataView`, `Int8Array`, `Uint8ClampedArray`,
+`ArrayBuffer`, `DataView`, `Int8Array`, `Uint8Array`, `Uint8ClampedArray`,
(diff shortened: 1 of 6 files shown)


ttmx:

protocol.md
```diff
@@ -168,9 +168,11 @@ The literal value `undefined`.
 
 The values Infinity, -Infinity, and NaN.
 
-`["bytes", base64]`
+`["bytes", base64]`, `["bytes", base64, type]`
 
-A `Uint8Array`, represented as a base64-encoded string.
+A `Uint8Array`, `ArrayBuffer`, `DataView`, or typed array, represented as a base64-encoded string.
```
kentonv: Let's include an explicit list of all supported type names here for completeness.

kentonv (commented):
Design looks good to me. I'm overloaded and don't have time to review line-by-line, maybe @dimitropoulos or @teamchong can take it.

protocol.md
```diff
@@ -168,9 +168,14 @@ The literal value `undefined`.
 
 The values Infinity, -Infinity, and NaN.
 
-`["bytes", base64]`
-
-A `Uint8Array`, represented as a base64-encoded string.
+`["bytes", base64]`, `["bytes", base64, type]`
+
+A byte container, represented as a base64-encoded string. If `type` is omitted, the receiver
+should deserialize bytes as its default `Uint8Array` for backwards compatibility. Otherwise,
```
ndisidore, line 174: Can you clarify endianness? Currently multi-byte typed arrays go on the wire as raw native-endian bytes: my recommendation is to mandate little-endian on the wire

src/serialize.ts
```diff
@@ -601,19 +649,45 @@ export class Evaluator {
...
               }
+            }
+            if (value.length === 2) {
               return bytes;
             }
+            if (typeof value[2] !== "string") {
+              throw new TypeError(`Unknown bytes type marker type: ${typeof value[2]}`);
+            }
+
+            let buffer = bytes.buffer.slice(bytes.byteOffset, bytes.byteOffset + bytes.byteLength);
```
ndisidore: I believe when misaligned untrusted input throws a raw RangeError ("byte length of Int16Array should be a multiple of 2"). We probably want to validate `byteLength % BYTES_PER_ELEMENT === 0`

src/serialize.ts
```diff
@@ -601,19 +649,45 @@ export class Evaluator {
...
+            switch (value[2]) {
```
ndisidore: This is likely more then a patch change
A current-main  receiver ignores `value[2]` entirely and returns a Uint8Array of raw bytes for a 3-element message (eg. new sender `Float64Array` → old receiver yields a 16-byte `Buffer`).

src/serialize.ts
```diff
@@ -73,6 +73,94 @@ export const DEFAULT_LIMITS: RpcLimits = {
...
+// big-endian hosts; it is exported solely so the behavior can be tested on
+// little-endian hosts.
+export function swapByteOrder(bytes: Uint8Array, elementSize: number): void {
+  if (elementSize !== 2 && elementSize !== 4 && elementSize !== 8) {
+    throw new RangeError(`Unsupported element size: ${elementSize}`);
+  }
+
+  let view = new DataView(bytes.buffer, bytes.byteOffset, bytes.byteLength);
+  for (let offset = 0; offset < bytes.byteLength; offset += elementSize) {
+    switch (elementSize) {
```
kentonv, line 150: I have a hunch that this would be significantly faster if you had the switch outside the for loop (with a separate for loop in each case).

However, nobody uses big endian anymore so I don't really care. :)

(I'm actually a bit shocked that JS uses host byte order instead of defaulting to little-endian...)

kentonv:
ughh Bonk accidentally pushed a commit. It does that sometimes and we don't know why. @ttmx can you force-push remove bonk's commit? Then I think we can merge this, no other changes needed.

src/serialize.ts
ttmx: That is how I had it implemented before :)

It's even faster if we use Buffer.swap* on platforms that support it, but after talking to Nathan we decided that since this is basically dead code it would be better to prioritize simplicity, and this has slightly less SLOCs :P

https://github.com/oven-sh/bun/issues/34231 here is part of an investigation I did on throughput

src/serialize.ts
```diff
@@ -738,27 +842,68 @@ export class Evaluator {
...
+            case "Uint8ClampedArray": return new Uint8ClampedArray(buffer);
+            case "Int16Array": return new Int16Array(buffer);
+            case "Uint16Array": return new Uint16Array(buffer);
+            case "Int32Array": return new Int32Array(buffer);
+            case "Uint32Array": return new Uint32Array(buffer);
+            case "BigInt64Array": return new BigInt64Array(buffer);
+            case "BigUint64Array": return new BigUint64Array(buffer);
+            case "Float32Array": return new Float32Array(buffer);
+            case "Float64Array": return new Float64Array(buffer);
+            default: marker satisfies never;
```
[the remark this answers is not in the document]
kentonv: The runtime check already happened on an earlier line.

kentonv (approved):

PR #209, opened 2026-07-07
Tidy the error-type map and WebSocket close-reason limit

ndisidore:
- serialize.ts: use a null-prototype object literal for ERROR_TYPES instead of Object.assign(Object.create(null), ...).
- websocket.ts: derive the close-reason cap from the RFC 6455 Close-frame size (125 - 2) rather than a bare 123, and shorten the comment.
- test: split a multi-byte character exactly at the limit so the streaming truncation path is actually exercised (the previous input landed on a character boundary).

No behavior change.

### DIFF
--- a/src/rpc.ts
+++ b/src/rpc.ts
@@ -205,13 +205,6 @@ class ImportTableEntry {
     //   a forwarded call. The caller is responsible for ensuring the last of these is handed off
     //   before direct calls can be delivered.
 
-    if (this.resolution) {
-      // Already resolved. A duplicate resolution would overwrite (and leak) the
-      // existing one, so dispose the incoming hook and ignore it.
-      resolution.dispose();
-      return;
-    }
-
     if (this.localRefcount == 0) {
       // Already disposed (canceled), so ignore the resolution and don't send a redundant release.
       resolution.dispose();
--- a/src/websocket.ts
+++ b/src/websocket.ts
@@ -7,6 +7,9 @@
 import { RpcStub } from "./core.js";
 import { RpcSession, RpcSessionOptions } from "./rpc.js";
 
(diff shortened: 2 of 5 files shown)


src/rpc.ts
```diff
@@ -205,13 +205,6 @@ class ImportTableEntry {
     //   a forwarded call. The caller is responsible for ensuring the last of these is handed off
     //   before direct calls can be delivered.
 
-    if (this.resolution) {
```
dimitropoulos, line 208: it'd be cool but not whatsoever required to add a comment here (is this the best place? yeah? right?) that explains that a path like this is not reachable since it's not super super clear otherwise that the case is as such.

src/websocket.ts
```diff
@@ -124,13 +124,13 @@ export class WebSocketTransport<T extends string | ArrayBuffer = string> {
     } else {
       message = `${reason}`;
     }
-    // WebSocket close reasons are limited to 123 UTF-8 bytes; a longer string makes
-    // close() throw. Truncate on a character boundary to stay within the limit.
+    // A Close frame is a control frame: its payload is <= 125 bytes (RFC 6455 §5.5), and the
+    // 2-byte status code leaves this for the UTF-8 reason. Longer reasons make close() throw.
+    // `stream: true` drops a trailing partial code point rather than emitting a replacement char.
+    const maxReasonBytes = 125 - 2;
```
dimitropoulos: since this is used/documented twice (here and in the tests above), please do a `export const MAX_REASON_BYTES = ....` somewhere in the library so that the value in the test and the value in the actual code can never drift by accident

dimitropoulos (approved):
some small improvements suggested, but not super blocking

PR #227, opened 2026-07-31
capnweb-validate: ignore extra args instead of refusing the call

teamchong:
`validateArgs` refused any call carrying more args than the method's declared parameter list. That breaks schema evolution: a caller built against a newer signature cannot pass a new parameter to a peer whose validator was generated from the old one, even though plain JS drops the unused arg.

Before:
```ts
// signature: greet(name: string)
stub.greet("bob", { locale: "en" });  // TypeError: refused Api.greet: no extra argument
```

After:
```ts
stub.greet("bob", { locale: "en" });  // extra arg ignored
```

Declared parameters are still validated, and a rest parameter still validates every extra arg.

### DIFF
--- a/.changeset/ignore-extra-args.md
+++ b/.changeset/ignore-extra-args.md
@@ -1,0 +1,5 @@
+---
+"capnweb-validate": patch
+---
+
+Ignore extra arguments past a method's declared parameters instead of refusing the call, and drop them before invoking the implementation.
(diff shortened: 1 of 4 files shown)


dimitropoulos (approved):
I'm not 100% sold on the premise - but I understand the reasoning.  maybe later it can be configurable (strict mode?)

dimitropoulos (changes requested):
let's document allowable evolution places (e.g. extra args, extra fields).

dimitropoulos (approved):

PR #229, opened 2026-08-05
Add support for sending `Set` over RPC

### LINKED ISSUE #221: support `Set` serialization type
implement support for Set to align with [Cloudflare Workers RPC capabilities](https://developers.cloudflare.com/workers/runtime-apis/rpc/#structured-clonable-types-and-more)

srikrsna:
Adds serialization support for `Set`. Doesn't special case `.map()` like we do for arrays, uses the default. I think we don't need it is because `Set` itself doesn't have one, so it is not expected as much as it is on an `Array`.

This does handle lazy values that are part of a set. Again, can't think of a use case but using a `Blob` as an element of `Set` is technically valid and also using an `RpcPromise` that you get back as part of the `Set`, so this implementation handles such cases. It handles it by not moving away from `LocatedPromise` which expects a parent and property accessor to replace the placeholder promise with the resolved value.

We could refactor the code to avoid the temporary setter but that felt like a bigger change, so once this and `Map` implementation merges we can revisit that.

If we want to revisit `.map` over set elements, I think we should block `.map` support for `Set`

Closes #221

### DIFF
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -27,6 +27,8 @@ let SERIALIZE_TEST_CASES: Record<string, unknown> = {
   '{"foo":[[123]]}': {foo: [123]},
   '{"foo":[[123]],"bar":[[456,789]]}': {foo: [123], bar: [456, 789]},
 
+  '["set",[1,2,"abc",[[123]]]]': new Set([1, 2, "abc", [123]]),
+
   '["bigint","123"]': 123n,
   '["date",1234]': new Date(1234),
   '["bytes","aGVsbG8h"]': new TextEncoder().encode("hello!"),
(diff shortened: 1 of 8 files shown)


srikrsna:

__tests__/index.test.ts
```diff
@@ -1573,6 +1575,117 @@ describe("promise pipelining", () => {
...
+    expect(result.strayProps).toStrictEqual([]);
+  });
+
+  it("collapses a resolution that equals an existing element", async () => {
+    await using harness = new TestHarness(new SetTarget());
+    let stub = harness.stub;
+    using promise = stub.square(3);
+
+    // Rebuilding the Set re-applies Set semantics, so the duplicate 9 drops out.
+    let result = await stub.inspect(new Set<unknown>([9, promise, "tail"]));
```
dimitropoulos, line 1656: this is a great test - I was thinking something along the lines of "what should happen if (perhaps in error!) over the wire someone sends `['set', [9, 9]` which will of course be handled by the `Set` constructor and return just `Set<[9]>`.  But my thought was that _maybe_ we want to detect when something was collapsed (perhaps cheaply by just checking input.length and output.length (deeply)).

but this test convinced me that, no, it's fine to just leave it to the `Set` constructor and it's not necessarily something bad to be worried about if something got [SameValueZero](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-samevaluezero)'d.

dimitropoulos (approved):

ndisidore (approved):

src/serialize.ts
```diff
@@ -841,6 +850,19 @@ export class Evaluator {
...
           }
           break;
+        case "set":
+          if (value.length === 2 && value[1] instanceof Array) {
+            let set = new Set();
+            let counter = 0;
+            for (let element of value[1]) {
+              let key = `${counter++}`;
+              let copy = this.evaluateImpl(element, set, key, depth + 1);
+              defineSetPromiseSlot(set, key, copy);
```
kentonv, line 860: Hmm. This is clever, but kinda hacky. I worry about it for a few reasons:

* Technically these property accessors will be visible to the app. You wouldn't expect apps to every try to access them, but they could, and I wonder if this will lead to any weirdness we aren't anticipating.
* This seems likely to be expensive. I imagine defining a property is not a particularly cheap operation, and we're defining one for every single member of the set.
* This is only needed to support a set of promises. But who would ever send a set of promises? That seems like a weird thing to do.

Maybe we could just ban promises in Sets?

But I suppose the same concerns apply to Maps. And in that case, it is a little bit more reasonable to imagine someone using promises as map values.

The way I had always imagined I'd handle that is that, I would pass the map key as the `property` in the recursive call to `evaluateImpl()`, and I would arrange that when we fill in promises later, if the parent object is a `Map`, we'd to `parent.set(key, value)` instead of `parent[key] = value`.

Granted, this approach is... more tedious to implement. But probably faster in the end.

(I suppose it could apply to sets too -- if the parent is a Set, insert into it, ignoring the key entirely.)

srikrsna: > Technically these property accessors will be visible to the app. You wouldn't expect apps to every try to access them, but they could, and I wonder if this will lead to any weirdness we aren't anticipating.

The only collision I see is if users also define properties. I went with a string accessor instead of a simple number index to reduce the chance of that. We can prefix the key with something like `@capnweb/` to almost eliminate this case.

> * This seems likely to be expensive. I imagine defining a property is not a particularly cheap operation, and we're defining one for every single member of the set.

It is only defined for elements that are promises. The function checks if the copy is an RpcPromise and only defines a property if it is. 

> * This is only needed to support a set of promises. But who would ever send a set of promises? That seems like a weird thing to do.

Not necessarily a set of promises, a set with partial promises is also possible: 
```typescript
    using promise = stub.square(3);
    let result = await stub.inspect(new Set([9, promise, 5])); // the other side will see Set([9, 5])
```
Again, can't think of a realistic scenario but reads alright to me. Supporting it means one less rule to remember. "It just works" is better DX.


> Maybe we could just ban promises in Sets?
> 
> But I suppose the same concerns apply to Maps. And in that case, it is a little bit more reasonable to imagine someone using promises as map values.

If we do ban promises in Sets, we should also ban promises in Map keys. For Map values, I think it will be odd not to support them.


> Granted, this approach is... more tedious to implement. But probably faster in the end.

In the current implementation, the cost applies only to Sets and Maps containing promises and doesn't impact other cases. If we agree that it is rare to use Maps and Sets, and rarer still for the elements to be promises, then it doesn't justify additional checks in every other case just to support these types. Not that the instance checks are costly but like you said implementing this is more tedious and spread out.

kentonv: > It is only defined for elements that are promises. The function checks if the copy is an RpcPromise and only defines a property if it is.

Oh, somehow I missed that, that resolves the performance concern I think.

> We can prefix the key with something like @capnweb/ to almost eliminate this case.

What about Symbols? I think you could just allocate new Symbols on the fly?

> If we do ban promises in Sets, we should also ban promises in Map keys. For Map values, I think it will be odd not to support them.

TBH we might want to be even more aggressive about map keys. It makes very little sense to use map keys that use identity-equality since there is no way to pass identity over the wire (since we intentionally don't support aliasing or cycles). So it might make sense to say that map keys can only be primitives.

I suppose you could argue that a promise might resolve to a primitive. But there's another problem there, which is what if you have multiple promise-keys in the same set or map which subsequently resolve to the _same_ primitive value? Awkward! So yeah I think we definitely should not support promises in sets or as map keys.

Agreed that for map _values_, we do need to support promises.

src/serialize.ts
```diff
@@ -323,7 +323,27 @@ export class Devaluator {
...
+              case "pipeline":
+              case "remap":
+                throw new TypeError(
+                    "Cannot serialize a promise as an element of a Set. Await the value before " +
+                    "adding it to the Set.");
+
+              case "export":
+              case "import":
+                throw new TypeError(
+                    "Cannot serialize a stub as an element of a Set.");
```
kentonv, line 342: Hmm, why?

Certainly I wouldn't generally recommend it but I don't see why it wouldn't work?

src/serialize.ts
```diff
@@ -522,6 +542,14 @@ export class Devaluator {
         // synchronously, and we MUST serialize the message synchronously. Hence, we have no choice
         // but to use streaming even for small blobs.
         let blob = value as Blob;
+
+        // The receiver must buffer the whole stream before it can construct the Blob, so it
+        // delivers the Blob as a promise. Set doesn't support a promise. To avoid the creation of a
+        // pipe, we reject it here.
+        if (parent instanceof Set) {
```
kentonv, line 549: Why not add "blob" to the switch under `case "set"` where all the other set-incompatible types are checked?

PR #232, opened 2026-08-06
Support `Map` over the RPC

### LINKED ISSUE #230: support `Map` serialization type
implement support for Map to align with [Cloudflare Workers RPC capabilities](https://developers.cloudflare.com/workers/runtime-apis/rpc/#structured-clonable-types-and-more)

srikrsna:
Adds serialization support for `Map`. Doesn't special case `.map()` like we do for arrays, uses the default. I think we don't need it is because `Map` itself doesn't have one, so it is not expected as much as it is on an Array. But I can see the use case of wanting to iterate over values of a `Map`.

Just like #229, we support lazy values for both keys and values of a map.

If we want to revisit `.map` over map elements, I think we should block `.map` support for `Map` in this.

Closes #230

### DIFF
--- a/.changeset/serialize-map.md
+++ b/.changeset/serialize-map.md
@@ -1,0 +1,5 @@
+---
+"capnweb": minor
+---
+
+Support serializing `Map` objects over RPC.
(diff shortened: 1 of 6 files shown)


teamchong:
recheck

PR #237, opened 2026-08-12
Share one RpcPromise alias between Result and the public export

Maximo-Guk:
Defines `RpcPromise` once in types.d.ts so `Result`'s arms and the public export resolve to the same alias instantiation. Fixes all TypeScript 7 (tsgo) depth errors (TS2321/TS2589) with no test assertion changes, and drops 5.9 type instantiations 175,019 → 151,466 — which also makes the `RpcSession<undefined>` workaround in index.test.ts unnecessary.


Maximo-Guk:

dimitropoulos (approved):

PR #239, opened 2026-08-12
Keep published runtime bundles ASCII-only

Maximo-Guk:
0.11.0's dist bundles carry a U+2212 (typographic minus) from a `websocket.ts` doc comment (#209). Discovered in [Cloudflare OS capnweb bump](https://github.com/cloudflare/cloudflare-os/pull/175/changes#diff-f0b1fee712e9299c02886b19b53ccff8b2c4c14f70b1179b79447cff90ba7d78R17). Consumers that inline the bundle - e.g. into a `data:` URL via `btoa()`, which only accepts Latin-1 - throw on it, and anything in U+0080–U+00FF would corrupt silently instead.

This fixes the comment and adds a post-build check that fails if any non-ASCII byte reaches a runtime bundle in `dist/`. Runtime strings that legitimately need non-ASCII can use `\u` escapes in source.

### DIFF
--- a/.changeset/ascii-only-dist.md
+++ b/.changeset/ascii-only-dist.md
@@ -2,4 +2,4 @@
 "capnweb": patch
 ---
 
-Keep the published runtime bundles ASCII-only. A doc comment introduced in 0.11.0 carried a U+2212 into every dist bundle, which breaks consumers that inline the bundle through Latin-1-only APIs like `btoa()`. The comment is fixed and the build now fails if any non-ASCII byte reaches a runtime bundle in `dist/`.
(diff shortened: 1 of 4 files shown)


.changeset/ascii-only-dist.md
```diff
@@ -0,0 +1,5 @@
+---
+"capnweb": patch
+---
+
+Keep the published runtime bundles ASCII-only. A doc comment introduced in 0.11.0 carried a U+2212 into every dist bundle, which breaks consumers that inline the bundle through Latin-1-only APIs like `btoa()`. The comment is fixed and the build now fails if any non-ASCII byte reaches a runtime bundle in `dist/`.
```
ndisidore: i think this was actually introduced several versions ago, cfos just went from 0.8.0 -> 0.11.0 so we just caught it

Maximo-Guk: IMG_1205

As far as I can tell, it seems to have gone out in the last release. Before this it looks like our last release was on July 6th, and the changes got merged on the 9th

teamchong (approved):

PR #241, opened 2026-08-12
fix: dispose call args on all failure paths per new StubHook ownership contract

ndisidore:
RPC call arguments (and map captures) leaked whenever a call failed before reaching a callee. A rejected pipeline promise, a broken or disposed stub, a failed argument serialization, or a sync throw inside a hook.

Practical example:
```ts
using api = newWebSocketRpcSession<Api>("wss://example.com/api");

let session = api.authenticate(token);       // NOT awaited — returns a promise stub
using uploader = session.getUploader();      // pipelined on the unresolved promise
uploader.write(chunk1, progressCallback);    // pipelined further, passes a capability
uploader.write(chunk2, progressCallback);
```
lets say token is expired so authenticate() rejects.  Every call pipelined behind it had already deep-copied its arguments, including a dup of `progressCallback's hook`. Before this PR, those copies were simply dropped on the rejection path: `progressCallback's` disposer never runs, its entry stays pinned.

Rather than patching each site with caller-side try/catch (as an earlier revision of this PR did), the fix defines the rule once on the abstract `StubHook`: `call()`, `stream()`, and `map()` take ownership of their args/captures even when they throw synchronously, and callers never dispose after invoking. 

The implementations that violated this are fixed
- `RpcImportHook.call()`/`stream()` (including mid-serialization failures, where disposal is safe because exported hooks are dups and the Devaluator rolls back its exports)
- -the default `stream()` (which leaked the result hook when `pull()` threw)
- `MapVariableHook`
- the `map` placeholder

`PromiseStubHook` now disposes its copied args only when its backing promise rejects i.e. where no callee ever existed to take them.

Also kept from the original PR: `PromiseStubHook.dispose()` chains on the backing promise so disposal stays ordered behind already-queued calls.

### DIFF
--- a/.changeset/promise-stubhook-disposal.md
+++ b/.changeset/promise-stubhook-disposal.md
@@ -1,0 +1,5 @@
+---
+"capnweb": patch
+---
+
+Fix `PromiseStubHook` to dispose copied call and stream arguments when the backing promise rejects, and to keep disposal ordered behind calls already queued on the promise.
--- a/.changeset/promise-stubhook-disposal.md
+++ b/.changeset/promise-stubhook-disposal.md
@@ -2,4 +2,4 @@
 "capnweb": patch
 ---
 
-Fix RPC argument and capture leaks on failure paths — rejected or broken destination hooks and local call errors now dispose the arguments they own — and keep `PromiseStubHook` disposal ordered behind already-queued calls.
+Establish and document the `StubHook` ownership contract: `call()`, `stream()`, and `map()` take ownership of their args/captures even when they throw synchronously, so callers must never dispose them after invoking. Fix implementations that violated the contract and leaked on failure paths: `RpcImportHook.call()/stream()` (including argument-serialization failures in `sendCall`/`sendStream`), `MapVariableHook`, the default `StubHook.stream()` (leaked the result hook when `pull()` threw), and the map-not-loaded placeholder. `PromiseStubHook` now disposes args only when its backing promise rejects — once resolved, the callee owns them — and its disposal remains ordered behind already-queued calls.
--- a/.changeset/promise-stubhook-disposal.md
+++ b/.changeset/promise-stubhook-disposal.md
@@ -2,4 +2,4 @@
 "capnweb": patch
 ---
 
-Fix RPC argument and capture leaks on failure paths: call arguments are now reliably disposed when a call is rejected, delivered to a broken or disposed stub, or fails to serialize. Internally, `StubHook.call()`/`stream()`/`map()` now document and follow a callee-takes-ownership contract, even on synchronous throw.
+Fix RPC argument and capture leaks on failure paths: call arguments are now reliably disposed when a call is rejected, delivered to a broken or disposed stub, or fails to serialize.
--- a/.changeset/promise-stubhook-disposal.md
+++ b/.changeset/promise-stubhook-disposal.md
@@ -2,4 +2,4 @@
 "capnweb": patch
 ---
 
-Establish and document the `StubHook` ownership contract: `call()`, `stream()`, and `map()` take ownership of their args/captures even when they throw synchronously, so callers must never dispose them after invoking. Fix implementations that violated the contract and leaked on failure paths: `RpcImportHook.call()/stream()` (including argument-serialization failures in `sendCall`/`sendStream`), `MapVariableHook`, the default `StubHook.stream()` (leaked the result hook when `pull()` threw), and the map-not-loaded placeholder. `PromiseStubHook` now disposes args only when its backing promise rejects — once resolved, the callee owns them — and its disposal remains ordered behind already-queued calls.
+Fix RPC argument and capture leaks on failure paths: call arguments are now reliably disposed when a call is rejected, delivered to a broken or disposed stub, or fails to serialize. Internally, `StubHook.call()`/`stream()`/`map()` now document and follow a callee-takes-ownership contract, even on synchronous throw.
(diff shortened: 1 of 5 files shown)


ndisidore:
One semantics question I hit while in here: if a call's args contain unresolved promises, `deliverCall()` waits on them before invoking the target, so a `dispose()` right after the call can win the race and the method runs on a disposed target. 

Pre-existing stuff: this PR just keeps disposal from overtaking *forwarding*, same as calling the destination hook directly. Is that intended, or should disposal also wait for in-flight deliveries? (That'd be a `deliverCall`/`TargetStubHook` lifetime change, so I didn't touch it.)

src/core.ts
```diff
@@ -2056,15 +2086,11 @@ export class PromiseStubHook extends StubHook {
...
-        hook.dispose();
-      }, err => {
-        // nothing to dispose
-      });
-    }
+    // Keep disposal behind calls already queued on this promise. Unlike pull(), dup(), and
+    // onBroken(), dispose() must not take a fast path through `this.resolution` even once it is
+    // available: it is the one destructive operation here, and a call chained on the promise just
+    // before disposal would otherwise be delivered after its target was disposed.
+    this.promise.then(hook => hook.dispose(), () => {});
```
ndisidore, line 2111: noting, but the one-microtask delay is invisible under dispose()'s already-asynchronous contract

this asymmetry is real but principled rather than leftover since dispose _requires_ e-order (and is the only destructive op)

src/core.ts
```diff
@@ -1992,18 +2001,39 @@ export class PromiseStubHook extends StubHook {
     // can't serialize them yet, we have to deep-copy them now.
     args.ensureDeepCopied();
 
-    return new PromiseStubHook(this.promise.then(hook => hook.call(path, args)));
+    return new PromiseStubHook(this.promise.then(
+        hook => {
+          try {
+            return hook.call(path, args);
```
kentonv: When `hook.call()` throws, is disposal of the args the caller's or callee's responsibility?

The changes to `ErrorStubHook` above seem to take it as the callee's responsibility, but this change here takes it as the caller's.

We should be consistent. I think it should always be the callee's responsibility.

src/core.ts
```diff
@@ -2056,15 +2086,11 @@ export class PromiseStubHook extends StubHook {
...
-      this.promise.then(hook => {
-        hook.dispose();
-      }, err => {
-        // nothing to dispose
-      });
-    }
+    // Keep disposal behind calls already queued on this promise. Unlike pull(), dup(), and
+    // onBroken(), dispose() must not take a fast path through `this.resolution` even once it is
+    // available: it is the one destructive operation here, and a call chained on the promise just
+    // before disposal would otherwise be delivered after its target was disposed.
```
kentonv: Comment is hard to understand. Maybe:

```suggestion
    // We can't take the fast path here when this.resolution is available because if the resolution
    // was just recently delivered, promise waiters may not have had a chance to consume it yet,
    // and disposing it would preempt them. If we always wait on `this.promise` then we'll
    // run after anything that was queued before us.
```

src/core.ts
ndisidore: I resolved this by establishing the ownership contract on the abstract `StubHook` itself: `call()`/`stream()`/`map()` take ownership of args/captures including on synchronous throw, so callers never dispose after invoking. 
`PromiseStubHook` now disposes args only on the rejection path, where no callee ever existed to take ownership.

Updated the PR body accordingly.
Let me know if you're happy with this

src/core.ts
```diff
@@ -173,10 +173,22 @@ function mapNotLoaded(): never {
 
 // map() is implemented in `map.ts`. We can't import it here because it would create an import
 // cycle, so instead we define two hook functions that map.ts will overwrite when it is imported.
-export let mapImpl: MapImpl = { applyMap: mapNotLoaded, sendMap: mapNotLoaded };
+export let mapImpl: MapImpl = {
```
kentonv: Please revert this diff. These stub functions always get replaced at startup time. Making them properly handle disposals is a waste of code.

.changeset/promise-stubhook-disposal.md
```diff
@@ -0,0 +1,5 @@
+---
+"capnweb": patch
+---
+
+Fix RPC argument and capture leaks on failure paths: call arguments are now reliably disposed when a call is rejected, delivered to a broken or disposed stub, or fails to serialize. Internally, `StubHook.call()`/`stream()`/`map()` now document and follow a callee-takes-ownership contract, even on synchronous throw.
```
kentonv: Implementation details don't belong in the public changelog.

```suggestion
Fix RPC argument and capture leaks on failure paths: call arguments are now reliably disposed when a call is rejected, delivered to a broken or disposed stub, or fails to serialize.
```

src/core.ts
```diff
@@ -1700,40 +1736,39 @@ abstract class ValueStubHook extends StubHook {
...
-        for (let cap of captures) {
-          cap.dispose();
-        }
-        throw err;
-      }
+      let {value, owner} = this.getValue();
+      let followResult = followPath(value, undefined, path, owner);
 
       if (followResult.hook) {
         return followResult.hook.map(followResult.remainingPath, captures, instructions);
```
kentonv, line 1774: If this throws, ownership of the captures has already been transferered, right? So it's incorrect for the catch block to dispose, below.

ndisidore: Good spot! You'd be right. Walked it back from the main catch

kentonv (approved):

PR #242, opened 2026-08-12
feat: construct RpcPromise from a Promise

ndisidore:
You can now write `new RpcPromise(promise)` (resolves the TODO). Calls made before the promise settles are queued and delivered in order once it does, and awaiting it yields the resolution. The promise can resolve to an `RpcTarget`, a stub, or a plain value.

@kentonv mentioned this re: reconnection in [cloudflare-os#172](https://github.com/cloudflare/cloudflare-os/pull/172#discussion_r5270434401)

Passing an existing `RpcPromise` adopts its hook directly rather than awaiting it. Two caveats are documented in the README: ownership of the resolution transfers to the `RpcPromise` (resolve with a `dup()` if you also keep the stub), and pending calls queue unboundedly (reject the promise when giving up).

Stacked because the new tests exercise rejection and dispose-before-resolution paths that only behave correctly with the `PromiseStubHook` fixes in #241. Landing that first keeps this diff feature-only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,10 +289,6 @@ In other words, this means:
 * If the promise rejects, the rejection propagates to all pipelined calls.
 * etc.
 
-Two special cases to be aware of:
-* Passing an existing `RpcPromise` (rather than a regular `Promise`) to the constructor adopts it directly, transferring ownership: the original promise is consumed and must not be used afterwards — use the wrapper in its place.
-* If a regular `Promise` resolves to an `RpcPromise`, JavaScript's promise machinery assimilates the thenable, which forces its resolution to be pulled. To keep a deferred capability lazy, resolve the promise with `rpcPromise.dup()` instead: `dup()` returns a non-thenable stub, which is adopted into the resolution and forwards pipelined calls without pulling.
-
 ### The magic `map()` method
 
 Every RPC promise has a special method `.map()` which can be used to remotely transform a value, without pulling it back locally. Here's an example:
--- a/README.md
+++ b/README.md
@@ -267,6 +267,25 @@ let profile = await api.getUserProfile(user.id);
 
 Whenever an `RpcPromise` is passed in the parameters to an RPC, or returned as part of the result, the promise will be replaced with its resolution before delivery to the receiving application. So, you can use an `RpcPromise<T>` anywhere where a `T` is required!
 
+#### Constructing `RpcPromise` from a `Promise`
+
+You can construct an `RpcPromise` yourself from a regular `Promise` (or any other thenable), using `new RpcPromise(promise)`. The result supports pipelining immediately: calls made before the promise settles are queued and delivered, in order, once it does, while awaiting it yields the promise's resolution. The promise may resolve to an `RpcTarget`, a stub, or a plain value; if it rejects, queued calls and `await`s fail with the rejection error, and `onRpcBroken()` callbacks are invoked.
+
+This is useful when the application knows a capability will exist but doesn't have it yet. For example, while re-establishing a broken session, you can publish a promise-backed stand-in, so that interim calls queue up and flow to the new connection once it is ready:
+
+```ts
+// reconnect() returns Promise<RpcStub<MyApi>>.
+let promise = new RpcPromise<MyApi>(reconnect());
+
+// Calls pipeline immediately, and are delivered once reconnect() resolves.
+let result = await promise.doSomething();
+```
+
+Two things to watch out for:
+
+* Ownership of the resolution transfers to the `RpcPromise`: disposing it disposes the target (or stub) that the promise resolved to. If you also want to keep the stub you resolved the promise with, resolve it with a `.dup()`.
+* Calls made while the promise is pending queue unboundedly, holding copies of their arguments. If the awaited capability may never arrive -- e.g. reconnection fails permanently -- reject the promise, so that queued calls fail rather than accumulate.
+
 ### The magic `map()` method
 
 Every RPC promise has a special method `.map()` which can be used to remotely transform a value, without pulling it back locally. Here's an example:
--- a/README.md
+++ b/README.md
@@ -289,6 +289,10 @@ In other words, this means:
 * If the promise rejects, the rejection propagates to all pipelined calls.
 * etc.
 
+Two special cases to be aware of:
+* Passing an existing `RpcPromise` (rather than a regular `Promise`) to the constructor adopts it directly, transferring ownership: the original promise is consumed and must not be used afterwards — use the wrapper in its place.
+* If a regular `Promise` resolves to an `RpcPromise`, JavaScript's promise machinery assimilates the thenable, which forces its resolution to be pulled. To keep a deferred capability lazy, resolve the promise with `rpcPromise.dup()` instead: `dup()` returns a non-thenable stub, which is adopted into the resolution and forwards pipelined calls without pulling.
+
 ### The magic `map()` method
 
 Every RPC promise has a special method `.map()` which can be used to remotely transform a value, without pulling it back locally. Here's an example:
--- a/README.md
+++ b/README.md
@@ -269,22 +269,25 @@ Whenever an `RpcPromise` is passed in the parameters to an RPC, or returned as p
 
 #### Constructing `RpcPromise` from a `Promise`
 
-You can construct an `RpcPromise` yourself from a regular `Promise` (or any other thenable), using `new RpcPromise(promise)`. The result supports pipelining immediately: calls made before the promise settles are queued and delivered, in order, once it does, while awaiting it yields the promise's resolution. The promise may resolve to an `RpcTarget`, a stub, or a plain value; if it rejects, queued calls and `await`s fail with the rejection error, and `onRpcBroken()` callbacks are invoked.
+You can construct an `RpcPromise<T>` directly from a regular `Promise<T>`, allowing you to perform promise pipelining on a regular local promise. Pipelined calls will wait until the inner promise resolves, then will be delivered, in-order, to the resolution. This is useful when you plan to obtain some stub in the future, but you want to allow code to start queuing calls on it immediately.
 
-This is useful when the application knows a capability will exist but doesn't have it yet. For example, while re-establishing a broken session, you can publish a promise-backed stand-in, so that interim calls queue up and flow to the new connection once it is ready:
+Wrapping a `Promise<T>` in this way is semantically identical to creating a local-loopback RPC and then invoking it. That is:
 
 ```ts
-// reconnect() returns Promise<RpcStub<MyApi>>.
-let promise = new RpcPromise<MyApi>(reconnect());
+// this...
+let rpcPromise = new RpcPromise(myPromise);
 
-// Calls pipeline immediately, and are delivered once reconnect() resolves.
-let result = await promise.doSomething();
+// is semantically the same as this...
+let rpcFunc = new RpcStub(() => myPromise);
+let rpcPromise = rpcFunc();
 ```
 
-Two things to watch out for:
-
-* Ownership of the resolution transfers to the `RpcPromise`: disposing it disposes the target (or stub) that the promise resolved to. If you also want to keep the stub you resolved the promise with, resolve it with a `.dup()`.
-* Calls made while the promise is pending queue unboundedly, holding copies of their arguments. If the awaited capability may never arrive -- e.g. reconnection fails permanently -- reject the promise, so that queued calls fail rather than accumulate.
+In other words, this means:
+* The result of the promise must be serializable.
+* If the promise resolution contains `RpcTarget`s or `Function`s, the `RpcPromise`'s resolution will replace them with stubs.
+* Ownership of any stubs in the Promise result is transferred away. If you want to keep your own copies, you need to `dup()` them.
+* If the promise rejects, the rejection propagates to all pipelined calls.
+* etc.
 
 ### The magic `map()` method
 
--- a/src/core.ts
+++ b/src/core.ts
@@ -607,14 +607,13 @@ export class RpcPromise extends RpcStub {
         // forward through the payload without forcing a pull, and a single-stub payload forwards
         // onBroken(), preserving brokenness.
         //
-        // A rejection is adopted as an ErrorStubHook rather than left to reject the backing
-        // promise. This way the promise chains backing queued calls never reject -- the calls
-        // land on the ErrorStubHook, which disposes their arguments -- and the error only
-        // surfaces through pull(), so a discarded pipelined call can't produce an unhandled
-        // rejection event.
+        // A rejection is left on the backing promise, like the rejection of a local async call:
+        // PromiseStubHook disposes the arguments of queued calls and surfaces the error through
+        // pull() and onBroken(). The ignoreUnhandledRejections() call keeps a never-used promise
+        // from firing an unhandled rejection event; a pipelined call whose result is neither
+        // awaited nor disposed still fires one, matching local async calls.
         let promiseHook = new PromiseStubHook(Promise.resolve(hook).then(
-            value => new PayloadStubHook(RpcPayload.fromAppReturn(value)),
-            err => new ErrorStubHook(err)));
+            value => new PayloadStubHook(RpcPayload.fromAppReturn(value))));
         promiseHook.ignoreUnhandledRejections();
         super(promiseHook, []);
       }
--- a/src/core.ts
+++ b/src/core.ts
@@ -583,9 +583,10 @@ export class RpcPromise extends RpcStub {
         // payload.
         let raw = unwrapStubAndPath(<RpcStub><unknown>hook);
         if (raw.pathIfPromise!.length > 0) {
-          // Property promise: get() returns an independent hook, and properties have no
-          // disposer, so there is nothing to neuter.
-          super(raw.hook.get(raw.pathIfPromise!), []);
+          // Property promise: share the source's hook and path, exactly like the source promise.
+          // The get() producing an independent hook happens lazily on first use, and properties
+          // have no disposer, so there is nothing to neuter.
+          super(raw.hook, raw.pathIfPromise);
         } else {
           let adopted = raw.hook;
           raw.hook = DISPOSED_HOOK;
@@ -1842,10 +1843,16 @@ abstract class ValueStubHook extends StubHook {
       let {value, owner} = this.getValue();
 
       if (path.length === 0 && owner === null) {
+        if (value instanceof Object && "then" in value) {
+          // The hook wraps a thenable (e.g. a workerd-native RpcPromise or RpcProperty), so it
+          // really does back a promise. get([]) asks for an independent hook aliasing the same
+          // promise, which is exactly dup().
+          return this.dup();
+        }
+
         // The only way this happens is if someone sends "pipeline" and references a
-        // TargetStubHook, but they shouldn't do that, because TargetStubHook never backs a
-        // promise, and a non-promise cannot be converted to a promise.
-        // TODO: Is this still correct for rpc-thenable?
+        // TargetStubHook wrapping a non-thenable, but they shouldn't do that, because such a
+        // hook never backs a promise, and a non-promise cannot be converted to a promise.
         throw new Error("Can't dup an RpcTarget stub as a promise.");
       }
 
@@ -2056,7 +2063,13 @@ class TargetStubHook extends ValueStubHook {
   }
 
   onBroken(callback: (error: any) => void): void {
-    // TODO: Should RpcTargets be able to implement onRpcBroken?
+    let target = this.target;
+    if (target && "then" in target) {
+      // The target is thenable (e.g. a workerd-native RpcPromise), so it backs a promise, which
+      // becomes broken if it rejects.
+      Promise.resolve(target).then(() => {}, callback);
+    }
+    // TODO: Should non-thenable RpcTargets be able to implement onRpcBroken?
   }
 }
 
--- a/src/core.ts
+++ b/src/core.ts
@@ -562,8 +562,8 @@ export class RpcStub extends RpcTarget {
 
 export class RpcPromise extends RpcStub {
   // Internally, an `RpcPromise` is constructed from a `StubHook` plus a property path. The
-  // application may instead pass a promise (or any other thenable) for the eventual resolution;
-  // calls made before it settles are queued and delivered, in order, once it does.
+  // application may instead pass a promise for the eventual resolution; calls made before it
+  // settles are queued and delivered, in order, once it does.
   constructor(hook: StubHook | PromiseLike<unknown>, pathIfPromise?: PropertyPath) {
     if (hook instanceof StubHook) {
       super(hook, pathIfPromise!);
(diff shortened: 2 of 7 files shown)


src/index.ts
```diff
@@ -58,10 +58,19 @@ export const RpcStub: {
...
+ * once it does, so an `RpcPromise` can stand in for a capability that doesn't exist yet -- for
+ * example, one that will only become available after a broken session has been re-established.
+ * Note that the `RpcPromise` takes ownership of the resolution: disposing it disposes the
+ * resolution, so resolve the promise with a `dup()` if you also intend to keep the stub.
  */
 export type RpcPromise<T extends RpcCompatible<T>> = RpcPromiseType<T>;
 export const RpcPromise: {
-  // Note: Cannot construct directly!
+  new <T extends Stubable>(value: PromiseLike<T | Stub<T>>): RpcPromise<T>;
+  new <T extends RpcCompatible<T>>(value: PromiseLike<T | Stub<T>>): RpcPromise<T>;
```
kentonv: I think these types may be redundant. Also as written I don't think it's possible for the type to be inferred, which means someone writing `new RpcPromise` always has to explicitly specify the `<T>`, but it would be nicer if they could just write `new RpcPromise(somePromise)` and have it inferred.

What if we define it as:

```ts
new <T extends RpcCompatible<T>>(value: Promise<T>): RpcPromise<Unstubify<T>>;
```

Does this work? Note that it means we no longer support non-Promise thenables but I think that's OK?

src/core.ts
```diff
@@ -585,6 +594,41 @@ export function unwrapStubTakingOwnership(stub: RpcStub): StubHook {
...
+  hook.ignoreUnhandledRejections();
+  return hook;
+}
+
+// Adopt the resolution of a promise passed to `new RpcPromise()` with "return" semantics, taking
+// ownership of any stubs within (including a stub as the root value). This is the same
+// representation used for the resolution of a local async call: pull() delivers the value,
+// pipelined calls forward through the payload without forcing a pull, and a single-stub payload
+// forwards onBroken(), preserving brokenness.
+function hookForResolution(value: unknown): StubHook {
```
kentonv: This one-liner function is only called in one place. Should be inline.

src/core.ts
```diff
@@ -585,6 +594,41 @@ export function unwrapStubTakingOwnership(stub: RpcStub): StubHook {
   }
 }
 
+// Given the application's argument to `new RpcPromise(value)`, produce the hook backing the
+// promise.
+function hookForPromiseArg(value: PromiseLike<unknown>): StubHook {
```
kentonv: This function is only called in one place and is not that long. I think the logic should just be inlined. It's easier to read that way.

src/core.ts
```diff
@@ -585,6 +594,41 @@ export function unwrapStubTakingOwnership(stub: RpcStub): StubHook {
...
+  let type = typeForRpc(value);
+  if (type === "stub" || type === "rpc-promise") {
```
kentonv: Pulling the `StubHook` out of a non-promise `RpcStub` and then wrapping it in an `RpcPromise`  doesn't necessarily work -- non-promise hooks may not know how to respond to `pull()`.

I think `RpcStub` will be handled correctly by the code below, though, so:

```suggestion
  if (type === "rpc-promise") {
```

README.md
```diff
@@ -267,6 +267,25 @@ let profile = await api.getUserProfile(user.id);
...
+let promise = new RpcPromise<MyApi>(reconnect());
+
+// Calls pipeline immediately, and are delivered once reconnect() resolves.
+let result = await promise.doSomething();
+```
+
+Two things to watch out for:
+
+* Ownership of the resolution transfers to the `RpcPromise`: disposing it disposes the target (or stub) that the promise resolved to. If you also want to keep the stub you resolved the promise with, resolve it with a `.dup()`.
+* Calls made while the promise is pending queue unboundedly, holding copies of their arguments. If the awaited capability may never arrive -- e.g. reconnection fails permanently -- reject the promise, so that queued calls fail rather than accumulate.
```
kentonv:
````suggestion
You can construct an `RpcPromise<T>` directly from a regular `Promise<T>`, allowing you to perform promise pipelining on a regular local promise. Pipelined calls will wait until the inner promise resolves, then will be delivered, in-order, to the resolution. This is useful when you plan to obtain some stub in the future, but you want to allow code to start queuing calls on it immediately.

Wrapping a `Promise<T>` in this way is semantically identical to creating a local-loopback RPC and then invoking it. That is:

```ts
// this...
let rpcPromise = new RpcPromise(myPromise);

// is semantically the same as this...
let rpcFunc = new RpcStub(() => myPromise);
let rpcPromise = rpcFunc();
```

In other words, this means:
* The result of the promise must be serializable.
* If the promise resolution contains `RpcTarget`s or `Function`s, the `RpcPromise`'s resolution will replace them with stubs.
* Ownership of any stubs in the Promise result is transferred away. If you want to keep your own copies, you need to `dup()` them.
* If the promise rejects, the rejection propagates to all pipelined calls.
* etc.
````

File: src/index.ts (hunk 1 above)
Comment (ndisidore) [reply src/index.ts]: It doesn't as written, but your instinct was right. collapsed the two constructor overloads into the single merged one: 
```ts
new <T extends RpcCompatible<T>>(value: Promise<T | Stub<T>>): RpcPromise<T>
```

you can see this tested here: https://github.com/cloudflare/capnweb/blob/33c01226c244fdbd9b3904dba2d50987a0cb7f71/__type-tests__/rpc-base-cases.test.ts#L154-L178

src/core.ts
ndisidore: done

src/core.ts
ndisidore: makes sense! integrated cleanly :heavy_check_mark:

README.md
ndisidore: applied

src/core.ts
ndisidore: goood catch! this was a legit bug

ndisidore:
@ask-bonk review this

src/index.ts
```diff
@@ -58,10 +58,19 @@ export const RpcStub: {
...
+ * invoking it: pipelined calls wait until the promise resolves, then are delivered, in order, to
+ * the resolution. This is useful when you plan to obtain some stub in the future, but want to
+ * allow code to start queuing calls on it immediately. Note that the `RpcPromise` takes
+ * ownership of the resolution: disposing it disposes the resolution, so resolve the promise
+ * with a `dup()` if you also intend to keep the stub.
  */
 export type RpcPromise<T extends RpcCompatible<T>> = RpcPromiseType<T>;
 export const RpcPromise: {
-  // Note: Cannot construct directly!
+  new <T extends RpcCompatible<T>>(value: Promise<T | Stub<T>>): RpcPromise<T>;
```
kentonv, line 73: I don't understand why we need `| Stub<T>` here.

Doesn't it block type inference? It seems like it would.

Either way though, I don't see why it would be needed.

ndisidore: > Doesn't it block type inference? It seems like it would.

It does not and there's actually a test for this already in `__type-tests__/rpc-base-cases.test.ts` around line 156

> Either way though, I don't see why it would be needed.

turns out is is needed
stubs aren't assignable to their target types (their methods return `Results`), so without `| Stub<T>` you can't pass a promise that resolves to a stub
without the union `new RpcPromise<TestTarget>(Promise.resolve(stub.dup()))` is a compile error.

per :robot: 
> The `| Stub<T>` member is what lets TypeScript match `Promise<Stub<PointTarget>>` against `Stub<T>` and pull out `T = PointTarget`

src/core.ts
```diff
@@ -561,9 +561,48 @@ export class RpcStub extends RpcTarget {
...
+  // settles are queued and delivered, in order, once it does.
+  constructor(hook: StubHook | PromiseLike<unknown>, pathIfPromise?: PropertyPath) {
+    if (hook instanceof StubHook) {
+      super(hook, pathIfPromise!);
+    } else {
+      if (pathIfPromise !== undefined) {
+        throw new TypeError("RpcPromise constructor expected one argument, received two.");
+      }
+
+      if (typeForRpc(hook) === "rpc-promise") {
```
ndisidore: hmmm

> On Cloudflare Workers, new `RpcPromise(env.SVC.someMethod())` — `typeForRpc` returns `'rpc-thenable'` (core.ts:150-154), not `'rpc-promise'`, so it takes the generic branch where `Promise.resolve()` immediately calls `.then()` on the workers promise

How do you want to handle the compatibility story here @kentonv

src/core.ts
```diff
@@ -561,9 +561,63 @@ export class RpcStub extends RpcTarget {
...
+      let kind = typeForRpc(hook);
+      if (kind === "rpc-promise") {
+        // Adopt an existing `RpcPromise` directly, transferring ownership of its hook: the source
+        // promise is neutered, as if disposed, and must not be used afterwards. In particular,
+        // adoption keeps the promise lazy -- assimilating it as a thenable would instead force its
+        // resolution to be pulled -- and preserves hook-local behavior such as brokenness. This
+        // applies only to promises, not bare stubs: a non-promise hook may not implement pull(),
+        // so a bare stub takes the generic path below, which adopts the stub into the resolution
+        // payload.
+        let raw = unwrapStubAndPath(<RpcStub><unknown>hook);
```
ndisidore, line 584: There was some discussion abut what happens when `new RpcPromise(existingRpcPromise)` adopts the source's hook. Opted for this consume-and-neuter approach (e.g. treated like a `move`) for consistency 

practically this means a user who wants copy semantics writes `new RpcPromise(source.dup())` themselves which is in line with convention for the rest of the lib

The alternative was something like
```ts
super(unwrapStubAndDup(<RpcStub><unknown>hook), []);
```
to auto-dup which would leave the original promise usable.

I'm impartial, but what is here seems marginally more consistent

kentonv: I think taking ownership is consistent with what happens with `Promise<T>` -- the `RpcPromise` will take responsibility for disposing the `T`.

Worth noting, though, that `dup()` won't work here: it returns an `RpcStub`, even when called on an `RpcPromise`.

src/index.ts
kentonv: It appears there are no test cases for `new RpcStub(promise)` where `promise` is type `Promise<RpcStub<T>>`.

This is what I'm confused about. In that case, is the result `RpcPromise<T>` or `RpcPromise<RpcStub<T>>`?

Note that these two types behave equivalently, but it seems like type deduction would have to choose one? Which does it choose?

I would actually lean towards `RpcPromise<RpcStub<T>>` being more correct -- and if that's the outcome we want, then I think then the `| Stub<T>` is not needed?

src/core.ts
```diff
@@ -561,9 +561,63 @@ export class RpcStub extends RpcTarget {
...
+        // onBroken(), preserving brokenness.
+        //
+        // A rejection is adopted as an ErrorStubHook rather than left to reject the backing
+        // promise. This way the promise chains backing queued calls never reject -- the calls
+        // land on the ErrorStubHook, which disposes their arguments -- and the error only
+        // surfaces through pull(), so a discarded pipelined call can't produce an unhandled
+        // rejection event.
+        let promiseHook = new PromiseStubHook(Promise.resolve(hook).then(
+            value => new PayloadStubHook(RpcPayload.fromAppReturn(value)),
+            err => new ErrorStubHook(err)));
```
kentonv: PromiseStubHook already handles the error case. We should not be handling it explicitly here.

From the comment above it sounds like this may have been done due to the missing disposal of arguments in that case -- but you've now fixed that.

src/index.ts
```diff
@@ -58,10 +58,28 @@ export const RpcStub: {
...
+ * invoking it: pipelined calls wait until the promise resolves, then are delivered, in order, to
+ * the resolution. This is useful when you plan to obtain some stub in the future, but want to
+ * allow code to start queuing calls on it immediately. Note that the `RpcPromise` takes
+ * ownership of the resolution: disposing it disposes the resolution, so resolve the promise
+ * with a `dup()` if you also intend to keep the stub.
+ *
+ * Passing an existing `RpcPromise` to the constructor adopts it directly, transferring ownership:
+ * the original promise is consumed (as if disposed) and the wrapper must be used in its place.
+ *
+ * Note that if a regular `Promise` resolves to an `RpcPromise`, JavaScript's promise machinery
```
kentonv: Delete this paragraph. What it's describing is not specific to the RpcPromise constructor -- it actually applies also when returning an RpcPromise from any async function or many other places. It's an interesting quirk but doesn't belong here.

Also the way the AI has described it here is probably inscrutable to most humans.

src/index.ts
```diff
@@ -58,10 +58,28 @@ export const RpcStub: {
...
+ * You may also construct an `RpcPromise` yourself from a regular `Promise`, using
+ * `new RpcPromise(promise)`, allowing you to perform promise pipelining on a local promise. This
+ * is semantically identical to creating a local-loopback RPC that returns the promise, and then
+ * invoking it: pipelined calls wait until the promise resolves, then are delivered, in order, to
+ * the resolution. This is useful when you plan to obtain some stub in the future, but want to
+ * allow code to start queuing calls on it immediately. Note that the `RpcPromise` takes
+ * ownership of the resolution: disposing it disposes the resolution, so resolve the promise
+ * with a `dup()` if you also intend to keep the stub.
+ *
+ * Passing an existing `RpcPromise` to the constructor adopts it directly, transferring ownership:
```
kentonv: I'd probably delete this paragraph too. It's describing a behavior that nobody would trigger intentionally -- if you know you have an RpcPromise, you would not bother constructing a new RpcPromise from it, you'd just use it as-is. So documenting the behavior is distracting.

README.md
```diff
@@ -267,6 +267,32 @@ let profile = await api.getUserProfile(user.id);
...
+```
+
+In other words, this means:
+* The result of the promise must be serializable.
+* If the promise resolution contains `RpcTarget`s or `Function`s, the `RpcPromise`'s resolution will replace them with stubs.
+* Ownership of any stubs in the Promise result is transferred away. If you want to keep your own copies, you need to `dup()` them.
+* If the promise rejects, the rejection propagates to all pipelined calls.
+* etc.
+
+Two special cases to be aware of:
```
kentonv: Delete these two special cases, nobody actually needs to be aware of them.

src/index.ts
ndisidore: Its not immediately obvious, but there is [a test that covers this case here](https://github.com/cloudflare/capnweb/blob/feat/from-promise/__type-tests__/rpc-base-cases.test.ts#L160-L162).

It chooses `RpcPromise<T>` and at runtime, you're right they behave equivalently, but the type system these aren't compatible.
Removing it would cause your own example in the readme to become a compile error and actually muddies up other types.

[Here's a ts playground link directly using capnweb types that illustrates behavior with (A) /without (B) stub better then I'll be able to explain with words :smile:  ](https://www.typescriptlang.org/play/?target=99&ts=6.0.3#code/PTAEGMEMAcDsHcCmAjUAFASqAxAJgCy4Bco8AFpAC6gDOZA9gK4A2AJqJWYqNI8swEtwoDNHBoATvQC2Amt3D1YNShMbhK9CbQEBzWFUYTuyRAH4AUCCthQoAIIBKULETxQAHgAqAPgAUAG6QzIyIJJIycojeoAA+ImIAypR83j4+jiSi4lKy8ml2oCAcXOhYfvHJqb6gjLACSo42hQBCzq7uaYHBoeG5UWmZCTmR+TWFE5OFxbD0tfVKNs1epdDMkACeulJ17JCMmgC0AGaIlOBcNBAwCCgAAgAMAHQAjC+vAORXvPxyXOyUDbQRA0J6gFbcSjwOZQWCsASsKiIZo0PQGFLGK6QYygaT0ViIZiIPZXAngdbGdiKZSqdSaCQ0AA0Olg4EhpVR+kMOLkpAolC+zV2iG0BOQSLBfjQfEE5I2zOyVWQBQAvKAlWknk0BNJoFpqABvYZebG6M7MwHA4ZKi1A7jZCJ5RC2q3ZADCMmgVAE-G4AF9QMdcqAAERQOBIZAhiwWcmQGhXND0ASwSgmiRm6iIAAelEQcKu2XTmdABosdhL2b8Q1gjGkpm0RuMGNgoAeoD95dx9ACiD8AH1WNmSLX6yLmYONiO6w2hgFk+wDZ3O2SKQolCoeMnU0qskkUsqkym06azj4YzM3J5fN0QmF0P0xnFrQe0hk9yMnQU7MVONxMKAFTqq+NR1A0sBNKu2LrjSwyOlE9gkGWdgdNeoA5nmBbDB6urer6b63r0D6jNEXjPoqIHpEMDqPqR56dpenQ3kEd59CRgwfvBYw+FMvHTGAszzOBFhQTi1KbjRJEtEhXaoTEGH5qwhZiDhXqUD6RIESxRFcXR1FiLpaQWJ2zSHGZhygO8oAAOr8qArD0CCoApqcxisgoDD0PIgZaKAkA8LRhz0MchwqHwZigOZUXRTFzT2C4WjSMEAgAF5OZoJSQqe1CWogADcoAtKQEgwFcf60AeHB2pKXiJPZiDQBIDRNepaVlVwzQGAA1sSoAAAZeH1zmtuVYFKMyNBzOVfUUdUPhDUl5xkDwKZXGRapHqmxZnFqsYbtQ-lqqhklOvYUq0U8mL0MwvZ+Hqx5Ko4TTidQqBHVeJ1RC050kZdILXbd907geT0xrloAAKIAI6MMEHgADIKjxap+F0zgqjxZEKVh8OgBFLygCQuDONjSnXv46OY+huaKYWeOWYToC4PTtLcCQxzBPIe2wRciDgF1-aQCQ0Ow8wHi5cFfkKgZtEeJtJ4ZmeyMcGoyIvRAXD8-2yDCzDcMS8coDINLn4DLNh7bgrmbpMrrMXmAUVM2CEIcNCtDAswgiwLoWI4gAcgA8mRx4ihckDe4gkC+pFMWx3HUXNJ9YzZTxvLkFQ6G9hIGwiGgbrXJ7oDNkYyiARgIIsJQHgYBkE1zAIArtcixQ0B7Xu6LU8jaIoBLOVc8AtZhfm6JAq2UE8Il82ugZ1Bo4Gd4g5c0JXd2cbL8vbZQ76gPOCIWIw8hLyvkDOMU9Bdfvh8V8wlB+Mgp9gBDGAYAHGBr+x5ty5bm-pH3flVVaWYnAUwd0JN5Rq+J1BOV8pAcAbJoCUBoPbGOZlQAAGYwQtC+EXa+1BYQfFetwL0CZerxj8q2DAeceDYkgNIamwI57e3-mFVAccLAhwkBzNkDhoACFLBwSAPV5arxfKkDeKc5wLg7JPOMYl9p+V4R+DU9heHnhgAIJ4lBBGIGEUDSgj0ihgHPoBcU8gvaICaOozR2jhEnz4oY0A58LBWK0UIy2d9nB8WKE-F+b97JMF9IcAeMBgTsAiAEBEIpq77lSE8OJv9BLMCUGabQdQWECGOAIEEyDHb4DBBDbMawhANwEYrSghxcp7FgEA70ShHGwGYDnaEEgur9wbsteK8c2HFD8OVBy4A6z5jzOwAkrlKTUMoHmCQsA2JOiePAdpS8AYihoNE8AGpfAUyeBAp0kEp7QQgPI0ZIpKS7mIl+T+4iynpAsMdGWJF7Bf2PD-PwxyJCnJBg4pxdzTbyBaE8raKdXmIDGcSAx3jn6vxyeZUAABWMEABZAQ2ZepXRCOpOpuUSDxVmBIJKgg2rMiKkSQRTchJ1K6eZGR091ayBRawWZAwrklniJc7+Kdzzq1oaAd67gk6IDOnS4kz15HIDobyuCtFvpCtYCKnmmsBa0N1qLcWdpJa0JNoZZlSseUq1CNzTcvMtZiuVfrNVhsxWavXuysp5EYkW2eRy22qtoWoIwHUdS0gTCIAoBE3yvJImpiEMEdC7SRSkE2Cgyl0borNBaPQTgK0PYpm4OHdgyAE3LUgPAUeOU5j+RYQVcqMMQQYtbLyXgxhGl8iEMtAAVLlOtzRyp8y8hsFQiA6GKF1F5JyCzOAT3jBsVkM9WRltoBQYENBqyli7JAJ4eJbovGZMTAqyAF09j7MupmTQ7Dq2zELFwM5w1qmzbmvyTxsxdn3TrI9Y5tCnpzSU9dV67DF2maAAA2mekpkBmQ-tesyA9QHkAAF0-JXBesZCwu92B0BgNkoAA)

kentonv: Yes, I understand that, without `| Stub<T>`, you would not be able to construct `RpcPromise<SomeTarget>` from `Promise<RpcStub<SomeTarget>>`.

But I'm not sure you should. I kind of think `RpcPromise<RpcStub<SomeTarget>>`  is the right answer here.

Consider: If you have an RPC method declared server-side like this:

```ts
  myMethod(): Promise<RpcStub<SomeTarget>>;
```

Then I believe the stubified type ends up being:

```ts
  myMethod(): RpcPromise<RpcStub<SomeTarget>>;
```

We do not elide the `RpcStub` here when we convert the return type in `RpcPromise` (at least, AFAICT looking at the code, though it's worth testing to verify).

The `RpcPromise` constructor should apply exactly the same transformation as would be applied to a method's return type.

ndisidore: ah, I do see what you're getting at and that makes sense  - the way its done now is inconsistent and even the readme example violates this.

however, `RpcPromise<RpcStub<T>>` types as a stub-of-a-stub where pipelined calls work, but you can't pass it as an RPC argument or use the awaited value as an `RpcStub<T>`.
[another small playground](https://www.typescriptlang.org/play/?target=99&ts=6.0.3#code/JYWwDg9gTgLgBAbzgJTAYwCoEMoHMCmMANCugMowCuARnAL5wBmUEIcARGlmAHYDu+auwDcAKFFoANlgDOMuGVb5seQnHwAPGPh4ATeakw4C8BKLhwsu3QAoNALjg9KIavigkAno+ev3ASh8XNyhEcws4KEJKKB44DTgAajhPMQs6UQzRXXwpHHw4PLk4AEEwYHUtHX1SI1VTcIB6RrgMPgg4PixPeRgOnLyojhgAC2B5EEIRiF1I6Nj5LELuLGpgSWAYT3Z7cQtmuABaAD5agAUWEHH8AB5FSZUTY8c+Eax4UYK0CB4ZGChKGg+qFcMAAG74eR8TYjOAAHwUVGoNwwxwAdHAAOrQADWMjR4TBwCwj0INkCcAurGudyUpJgxzScAOJ3OlxphgoNFpD2MhGOz06b3gmwA5LNQRDOjCIJQYBiAKJQFhQeRuSQQPgEixErBc6jkxxUq4yW6cpE85R8hmMvZwSimmxgRzm7n3K31Y4U3whMRZAbSIbfX7wbjAF3kC1lYC20QHXX08lwP40YCMYCQuB9NnU02W+nouMtXX6pMWFNrdOZ7OGY0cyNuunWgVHQ5ZkYFV20cZwAByAHkMOoNjldATg38mJd6XAALyWcpohPW8liCfwZisfVzhfAJfE0v+MRFymyGTAHi4dsFMDs02WRZwcpgfAbHj4WbIM4AYUseBcOgwLsYZog6+A2JuID0kezItP2ADSoggWBEGXPqMEHAqyDIP2yA+BAoZyMAuA8KskgFNmnxPjgWCTNooRbC+4gHCUXSbBeV74FgaAjMBMieDwaBMJQgkwMAPxwDkIAQEmZgWOulg7lgbEbtO1pMgpWD9jiEZoPq+bNkpwiwXACGtnMMgQJIEK9B0SxgNIF7JkiJCyNekQiWJkySRAkJ2gptDzspWCbFOW5IhpPyTtQABCVi6fp7oFju1DGZh2G4Y4uiytQ5GHBWaYZrMMhvC+JDUFAWB6HAJogO8PF2gcZzAC+b4fsskiSPIfzrJIcClS+cTZu0UA4nAEnUARIwkK8wA8XAPavJ47m6GmjDuDoaAFD2XH8VmHQmjIuzyVF8A8AAjEEfihEFKlhVB1polYtjnSQABMGFwYhJ0hk4b1XSESl3ZB+pPdYNivXAH1pV9mRAA)

perhaps we can make `Result<Stub<T>>` elide to `RpcPromise<T>`? eh

ndisidore: I missed that this style of declaration is actually broken on main outside of this branch anyway right now.
Which is to say, on main a method declared like `viaStub(): Promise<RpcStub<SomeTarget>>` stubifies to `RpcPromise<RpcStub<SomeTarget>>`, and that type can't be passed as a pipelined RPC argument and its awaited value isn't assignable to `RpcStub<SomeTarget>`

So this really just exposed an existing issue a bit more because the constructor forced a decision about which of the two spellings is canonical.

I started walking this back and claude pushed back saying we would be giving up the "flagship feature" of this work, which I find a bit editorial but it gave me pause

dimitropoulos: We spent a two hours on this today (@ndisidore and @teamchong and I) and looked at the problem from a few directions, and it seems like this is an underlying issue in the first place that was only surfaced once we started exposing promises as arguments that can be stubbed.  We have a working solution for allowing `Result<T>` to Unstubify T - and that seems to work.

From where I'm standing, it seems like `| Stub<T>` was required all along.

---

In the future (I'm thinking, early October, right now) we need to do something like the native `Awaited` type to recursively unpack things _so that_ we don't need to add the `Stub<T>` union since, like was discussed above, it shouldn't really matter since they should all act sorta the same no matter how many layers it's wrapped.

kentonv (approved):

src/core.ts
ndisidore: Agreed - removed as it seems more consistent that way
a pipelined call on a rejected promise whose result is leaked (never awaited or disposed) now fires an unhandled-rejection event, matching what local async calls already do

PR #245, opened 2026-08-12
fix: create a fresh error per stub disposal for better stack traces

ndisidore:
## Motivation

While investigating production Sentry reports of `Error: Attempted to use RPC stub after it has been disposed.`, we found the error impossible to debug: every occurrence shares a single module-level `Error` object (`DISPOSED_HOOK`), so its stack trace points at capnweb's module initialization rather than anywhere useful. It also collapses all Sentry events into one group regardless of cause.

## Change

Delete the `DISPOSED_HOOK` singleton and create the `ErrorStubHook` (and its `Error`) fresh inside the `Symbol.dispose` disposer. The stack trace now points at the disposal site — the thing you actually need to see when tracking down a use-after-dispose.

- Message text is byte-identical, so existing Sentry grouping/alerting on the message keeps working.
- Allocation cost is one `Error` per explicit dispose — negligible.
- Double-dispose keeps today's semantics (the latest dispose site wins).

## Testing

- New tests: calling a disposed stub rejects with the expected message, and two separate disposals produce distinct error objects (fails against the singleton implementation).
- `npm run build`, `npx vitest run --project node --project workerd` (547 passing), `npm run test:types` all green.
- Manual sanity check against `dist/`: the rejection's stack names the function that disposed the stub.

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -335,9 +335,6 @@ export class ErrorStubHook extends StubHook {
   }
 };
 
-const DISPOSED_HOOK: StubHook = new ErrorStubHook(
-    new Error("Attempted to use RPC stub after it has been disposed."));
-
 // A call interceptor can be used to intercept all RPC stub invocations within some synchronous
 // scope. This is used to implement record/replay
 type CallInterceptor = (hook: StubHook, path: PropertyPath, params: RpcPayload) => StubHook;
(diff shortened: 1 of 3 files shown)


src/core.ts
```diff
@@ -391,7 +388,10 @@ const PROXY_HANDLERS: ProxyHandler<{raw: RpcStub}> = {
       // We only advertise Symbol.dispose on stubs and root promises, not properties.
       return () => {
         stub.hook.dispose();
-        stub.hook = DISPOSED_HOOK;
+        // Create the error here rather than sharing a singleton so that its stack trace points
+        // at the disposal site, which is what you need to see when debugging a use-after-dispose.
+        stub.hook = new ErrorStubHook(
```
kentonv, line 393: I guess you closed this PR so maybe it's no longer relevant, but:

This is too expensive. It means every disposal will allocate several objects and acquire a stack trace.

PR #246, opened 2026-08-13
Document TypeScript lib requirements

### LINKED ISSUE #112: `RpcStub<T>` doesn't preserve object return types - only primitives work
## Description

When using `newWebSocketRpcSession` or `newHttpBatchRpcSession` with a TypeScript generic parameter, the `RpcStub<T>` type preserves primitive return types (string, number, boolean) but fails to preserve object/interface return types, returning `any` instead.

## Minimal Reproducible Example

```typescript
import { newWebSocketRpcSession, type RpcStub } from 'capnweb';

// Define types
type ApiMessage = { message: string; timestamp: string };

// Define API interface
interface MyApi {
  getPrimitive(): Promise<string>;           // Returns primitive
  getObject(): Promise<ApiMessage>;          // Returns object
  hello(name: string): Promise<ApiMessage>;  // Returns object
}

// Create client
const apiStub = newWebSocketRpcSession<MyApi>('wss://example.com/api');

// Primitives work correctly ✅
const primitive = await apiStub.getPrimitive();
//    ^? Type is 'string' ✅

// Objects return 'any' ❌
const object = await apiStub.getObject();
//    ^? Type is 'any' ❌

const helloResult = await apiStub.hello('world');
//    ^? Type is 'any' ❌
```

## Expected Behavior

All method return types should be preserved, regardless of whether they're primitives or objects:

```typescript
const primitive = await apiStub.getPrimitive();
//    ^? string ✅

const object = await apiStub.getObject();
//    ^? Should be: ApiMessage

const helloResult = await apiStub.hello('world');
//    ^? Should be: ApiMessage
```

## Actual Behavior

Only primitive return types are preserved. Object/interface return types become `any`:

```typescript
const apiStub: RpcStub<MyApi> = newWebSocketRpcSession<MyApi>('wss://example.com/api');
//    ^? Correctly shows: RpcStub<MyApi>

const primitive = await apiStub.getPrimitive();
//    ^? string ✅ Works!

const object = await apiStub.getObject();
//    ^? any ❌ Should be ApiMessage

const helloResult = await apiStub.hello('world');
//    ^? any ❌ Should be ApiMessage

// No type checking for objects
console.log(helloResult.anythingGoes); // No error
```

## Root Cause

The `RpcStub<T>` type definition appears to handle primitives correctly but wraps object return types in complex proxy types (for promise pipelining support) that resolve to `any` instead of the original type.

## Workaround

Manually cast each result:

```typescript
const helloResult = await apiStub.hello('world') as ApiMessage;
//    ^? Now correctly typed as: ApiMessage ✅
```

## Environment

- **capnweb version**: 0.2.0
- **TypeScript version**: 5.8.3
- **Node version**: 22.15.1

ting-hong-shieh:
## Summary

- document the TypeScript target and library definitions required by Cap'n Web's public types
- explain which definitions cover web APIs, Promise.withResolvers(), and explicit resource management
- update the browser and Workers examples to use the documented libraries

## Testing

- npm run build
- npm run test:types
- TypeScript compile for examples/worker-react/client
- TypeScript compile for examples/worker-react/server
- npm run build in examples/worker-react/client
- npm test (34 files, 1,171 tests across Node, Workers, Chromium, Firefox, and WebKit)
- standalone type check with the documented minimum libraries, including an assertion that object results do not resolve to any

Closes #112

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,29 @@ Cap'n Web is more expressive than almost every other RPC system, because it impl
 npm i capnweb
 ```
 
+### TypeScript configuration
+
+Cap'n Web's type definitions refer to several recent built-in APIs. If your project sets
+[`compilerOptions.lib`](https://www.typescriptlang.org/tsconfig/#lib) explicitly, use this as a
+baseline:
+
+```json
+{
+  "compilerOptions": {
+    "target": "ES2022",
+    "lib": ["ES2022", "DOM", "ES2024.Promise", "ESNext.Disposable"]
+  }
+}
+```
+
+`ES2024.Promise` provides the types for `Promise.withResolvers()`, while `ESNext.Disposable`
+provides `Disposable`, `Symbol.dispose`, and `Symbol.asyncDispose`. Cap'n Web supplies runtime
+polyfills for those APIs, so the emit target can remain `ES2022`. `DOM` provides the web platform
+types exposed by the main package.
+
+If your project already includes `ESNext`, it covers the Promise and disposable definitions above.
+Keep any additional libraries required by your environment, such as `DOM.Iterable` or `WebWorker`.
+
 ## Example
 
 A client looks like this:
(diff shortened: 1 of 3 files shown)


ting-hong-shieh:

kentonv:
Hi @ting-hong-shieh,

Thanks for trying to help.

PR #247, opened 2026-08-13
Add bidirectional WebSocket example

### LINKED ISSUE #30: clarify how bidirectional communication should work

ting-hong-shieh:
## Summary

- add a runnable Node client and server that each expose a root RPC interface over one WebSocket
- show the server calling the client's showNotification() method before returning the original greet() call
- connect the WebSocket client guide to the example and point retained method callbacks to the existing .dup() lifecycle guidance

## Testing

- npm run build
- npm run test:types
- npm test (34 files, 1,171 tests across Node, Workers, Chromium, Firefox, and WebKit)
- node --check for both example programs
- end-to-end server and client run with the documented notification and response output

Closes #30

### DIFF
--- a/README.md
+++ b/README.md
@@ -533,6 +533,9 @@ console.log(await stub.greet("Bob"));
 // response back over the wire.
 ```
 
+For a complete example that exposes an RPC interface on both ends of the same connection, see
+[Bidirectional WebSocket calls](examples/websocket-bidirectional/README.md).
+
 ### HTTP server on Cloudflare Workers
 
 The helper function `newWorkersRpcResponse()` makes it easy to implement an HTTP server that accepts both the HTTP batch and WebSocket APIs at once:
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,6 +1,7 @@
 Examples
 
 - batch-pipelining: Node server + client. Shows batching and pipelining to execute a dependent sequence of RPC calls in a single HTTP round trip, with timing vs sequential.
+- websocket-bidirectional: Node server + client. Shows each side exposing an RPC interface over the same WebSocket connection.
 - worker-react: Cloudflare Worker backend + React frontend. Shows the same pattern from a browser app, served by the Worker.
 
 Notes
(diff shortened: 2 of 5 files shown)


ting-hong-shieh:

kentonv:
Hi @ting-hong-shieh,

Thanks for trying to help.

PR #248, opened 2026-08-13
feat(rpc): make onRpcBroken registrations disposable

### LINKED ISSUE #210: onRpcBroken callbacks are retained for the session's lifetime after their import is disposed (and fire at teardown)
**Version:** capnweb 0.10.0 (tag `92baaed`), Node 24.

**Summary:** an `onRpcBroken` callback registered on an import entry whose `resolution` is unset — which includes ordinary remote capability stubs, not just pending promises — is retained in the session-level `RpcSessionImpl.onBrokenCallbacks` list after the capability is fully disposed and released. The callback then fires at session teardown with the teardown error, long after the capability it observed was cleanly disposed. Since session abort doesn't clear the list either, the closures stay reachable for as long as the `RpcSessionImpl` remains reachable.

**Mechanism** (at the tag, `src/rpc.ts`):

- `ImportTableEntry.onBroken()` pushes the callback onto `session.onBrokenCallbacks` and records the index in the entry's `onBrokenRegistrations`.
- On final disposal of an entry with no resolution, `dispose()` → `abort()` sets `this.onBrokenRegistrations = undefined` **without deleting the corresponding entries from `session.onBrokenCallbacks`**, then `sendRelease()` removes the entry.
- This is asymmetric with `resolve()`, which explicitly reconciles the registrations: it deletes the old session slot when the callback moves elsewhere, or removes a redundant newly-added same-session slot while preserving the original ordering. The disposal path performs no equivalent cleanup.
- The comment in `abort()` ("The RpcSession itself will have called all our callbacks") holds when the *session* aborts the entry, but not when local `dispose()` calls `abort()` first.

**Observable consequences:**

1. One strong callback reference (plus anything the closure strongly captures) accumulates per registration for as long as the `RpcSessionImpl` is reachable — unbounded on a long-lived session with capability churn.
2. Callbacks belonging to cleanly-disposed capabilities still fire at session teardown, and with the teardown error rather than any disposal-related reason.
3. The retention is invisible to `getStats()`, which reports only the import/export tables — those return to baseline.

**Reproduction** (self-contained, `node repro.mjs` in a project with capnweb 0.10.0 installed):

```js
const { RpcSession, RpcTarget } = await import("capnweb");

// Trivial in-memory transport pair.
function transportPair() {
  const mk = () => ({ queue: [], waiter: null, closed: null });
  const a = mk(), b = mk();
  const side = (inbox, outbox) => ({
    send(message) {
      if (outbox.waiter) { outbox.waiter.resolve(message); outbox.waiter = null; }
      else outbox.queue.push(message);
    },
    async receive() {
      if (inbox.queue.length > 0) return inbox.queue.shift();
      if (inbox.closed) throw inbox.closed;
      inbox.waiter = Promise.withResolvers();
      return inbox.waiter.promise;
    },
    abort(reason) {
      inbox.closed = reason;
      if (inbox.waiter) { inbox.waiter.reject(reason); inbox.waiter = null; }
    },
    breakNow(reason) { this.abort(reason); },
  });
  return [side(a, b), side(b, a)];
}

class Thing extends RpcTarget { poke() { return "ok"; } }
class Main extends RpcTarget { getThing() { return new Thing(); } }

const [clientTransport, serverTransport] = transportPair();
void new RpcSession(serverTransport, new Main());
const client = new RpcSession(clientTransport);
const main = client.getRemoteMain();

console.log("baseline stats:", client.getStats());

let disposedFires = 0;
for (let i = 0; i < 6; i += 1) {
  const thing = await main.getThing();   // fresh capability, one handle
  thing.onRpcBroken(() => { disposedFires += 1; });
  await thing.poke();
  thing[Symbol.dispose]();               // fully released
}
await new Promise((r) => setTimeout(r, 50));
console.log("stats after 6 acquire/dispose cycles:", client.getStats()); // back to baseline
console.log("fires while session healthy:", disposedFires);              // 0

clientTransport.breakNow(new Error("connection lost"));
await new Promise((r) => setTimeout(r, 100));
console.log("fires at session death, all from disposed imports:", disposedFires); // 6
```

Observed output: `getStats()` returns to `{ imports: 1, exports: 1 }` after the six acquire/dispose cycles, zero callbacks fire while the session is healthy, and **all six** fire when the transport later breaks.

**Why this looks like a bug rather than intent:** the README describes `onRpcBroken` as monitoring a stub's "brokenness" and says the callback runs if something causes all further method calls and property accesses on that stub to throw. The follows-the-resolution behavior (which the test suite pins, and which `resolve()`'s session-array reconciliation supports) stops applying once the capability is fully released. Either defensible disposal semantic — unregister on disposal, or treat disposal as breakage and fire immediately with the disposal reason — differs from the current behavior of staying silent at disposal and firing much later with the later session-abort error. No test or example appears to pin the current post-disposal late fire.

**Fix sketch:** mirror `resolve()`'s cleanup on final disposal — delete the entry's registrations from `session.onBrokenCallbacks` before discarding `onBrokenRegistrations`. One care point: the deletion should distinguish healthy local disposal from session-abort reentrancy (a callback that disposes another entry mid-iteration must not delete callbacks that haven't fired yet), and existing tests pin global registration order across entries.

**Context:** found while building a subscription surface (client-passed observer capabilities, one `onRpcBroken` per subscription for reaping) on a long-lived Unix-socket session. Our registrations are bounded per connection and the callbacks idempotent, so this isn't urgent for us — filing for correctness. Checked for prior reports: #153/#168 (receive-loop retention, fixed in 0.10.0), #154, #151, and #172 are adjacent but concern different lifecycles.

### LINKED ISSUE #234: onRpcBroken should return a Disposable so registrations can be undone
**Version:** capnweb 0.10.0. Applies to `RpcStub.onRpcBroken` / `RpcPromise.onRpcBroken`.

## Summary

`onRpcBroken()` returns `void`, so a registration can never be undone. Registering the same callback twice fires it twice, and there is no `offRpcBroken`. I would like it to return a `Disposable` handle that unregisters the callback.

## Why this matters

Every other subscription-shaped API in the ecosystem hands back something you can use to stop listening: `addEventListener`/`removeEventListener`, `AbortSignal`, or a returned unsubscribe function. `onRpcBroken` is the odd one out, and the places you naturally want to call it are exactly the places where things get re-run.

The clearest case is a UI framework. A React effect that registers a handler and re-runs (a dependency changes, StrictMode double-invokes, a component remounts against a session that outlives it) has no way to clean up after itself, so handlers accumulate on a long-lived stub and each break fires all of them. The workaround is to hoist registration to the point where the session is created, which works but constrains where the code can live, and pushes people toward a mutable "current handler" variable that the permanent callback dispatches through.

Our own documentation currently has to warn readers about this rather than show them the fix:

> **`onRpcBroken` cannot be unregistered.** It returns nothing, and registering twice on the same stub fires twice. Register it where the session is created rather than in an effect that might re-run.

## Reproduction

```js
const ret = main.onRpcBroken(() => {});
console.log(ret);                    // undefined

let fires = 0;
const cb = () => { fires++; };
main.onRpcBroken(cb);
main.onRpcBroken(cb);                // same function, registered twice
transport.breakNow(new Error('connection lost'));
// fires === 2
```

Deduplicating by function identity would not be a fix on its own, since two independent subscribers legitimately registering the same function should each be honoured. What is missing is a handle.

## Proposed solution

Return a `Disposable`:

```ts
interface StubBase<T = unknown> extends Disposable {
  onRpcBroken(callback: (error: any) => void): Disposable;
}
```

```js
const subscription = stub.onRpcBroken(handleBroken);
// ...later
subscription[Symbol.dispose]();
```

which makes the ergonomic form work directly, and matches how the rest of the library already asks you to think about lifetimes:

```js
{
  using subscription = stub.onRpcBroken(handleBroken);
  await doWorkThatMightBreak();
}   // unregistered here
```

Notes on the shape:

- **Non-breaking.** The method currently returns `undefined`, so existing call sites that ignore the result keep working unchanged.
- **Idempotent.** Disposing twice, or disposing after the callback has already fired, should be a no-op rather than an error.
- **Disposing the stub** should continue to release registrations as it does today; the handle is for the narrower case of ending one subscription early.
- Returning a bare `() => void` unsubscribe function is the other obvious option, but `Disposable` composes with `using` and is the convention the library has already committed to for stubs and for the revoker objects people build on top of `Symbol.dispose`.

## Implementation sketch

The bookkeeping this needs is largely present. `RpcImportHook.onBroken()` (`src/rpc.ts`) already pushes onto the session-level array and records the slot index:

```ts
let index = this.session.onBrokenCallbacks.length;
this.session.onBrokenCallbacks.push(callback);
if (!this.onBrokenRegistrations) this.onBrokenRegistrations = [];
this.onBrokenRegistrations.push(index);
```

and `resolve()` already deletes individual slots with `delete this.session.onBrokenCallbacks[i]`, preserving the ordering that existing tests pin. A returned handle would close over the hook and that index and perform the same deletion.

The abstract `StubHook.onBroken` (`src/core.ts:315`) and its implementations would change return type. Two are trivial (`ErrorStubHook` fires immediately, so it can return an already-disposed no-op handle; the `RpcTarget` hook at `src/core.ts:1967` is currently a no-op). `PromiseStubHook.onBroken` forwards to a resolution that may not exist yet, so its handle needs to unregister from whichever hook it eventually forwarded to.

## Relationship to #210

#210 reports that registrations on a disposed import are retained for the session's lifetime and fire at teardown. That is a bug in the same bookkeeping this feature would build on, and its fix sketch (mirroring `resolve()`'s cleanup on the disposal path) is roughly the operation a disposable handle needs to perform on demand. It probably makes sense to do #210 first, or to do both together.

## Alternative worth considering alongside

An `AbortSignal` option would compose well with code that already has one for teardown:

```js
stub.onRpcBroken(handleBroken, { signal: controller.signal });
```

Worth noting for anyone who finds this issue by searching: this would be a purely local API, so it is unaffected by the fact that `AbortSignal` cannot currently be serialized and sent over the wire.

## One reason to settle this soon

`onRpcBroken` does not exist in the Workers Runtime's native RPC yet, and the two systems are converging deliberately. Fixing the signature while Cap'n Web is the only implementation avoids having to change it later in two places, one of which would need a compatibility flag.

ting-hong-shieh:
## Summary

- return an idempotent `Disposable` from `onRpcBroken()`
- unregister callbacks when their handle or owning stub is disposed, including unresolved promise and import hooks
- preserve callback order when a callback disposes a later stub during session abort
- update the public and validation types, README, tests, and changeset

## Testing

- `npm run build`
- `npm run test:types`
- `npm test -- --project=node __tests__/index.test.ts`
- the new tests pass in Node, workerd, Chromium, Firefox, and WebKit
- `npm test`: 1,188 of 1,189 tests pass locally; the existing WebKit `applies backpressure when window fills up` test times out in the full file on both this branch and a clean `origin/main`, while passing in isolation

Closes #234
Closes #210

### DIFF
--- a/.changeset/calm-dodos-listen.md
+++ b/.changeset/calm-dodos-listen.md
@@ -1,0 +1,5 @@
+---
+"capnweb": minor
+---
+
+Return a disposable registration from `onRpcBroken()` so callers can stop listening without disposing the stub. Disposing a stub now also removes its registered callbacks.
--- a/README.md
+++ b/README.md
@@ -421,13 +421,17 @@ Note that if you pass the same `RpcTarget` instance to RPC multiple times -- thu
 
 ### Listening for disconnect
 
-You can monitor any stub for "brokenness" with its `onRpcBroken()` method:
+You can monitor any stub for "brokenness" with its `onRpcBroken()` method. It returns a
+`Disposable` registration, which lets you stop listening without disposing the stub:
 
 ```ts
-stub.onRpcBroken((error: any) => {
+using registration = stub.onRpcBroken((error: any) => {
   console.error(error);
 });
 ```
+
+Calling `registration[Symbol.dispose]()` more than once is safe. Disposing the stub also removes
+its registrations without calling them.
 
 If anything happens to the stub that would cause all further method calls and property accesses to throw exceptions, then the callback will be called. In particular, this happens if:
 * The stub's underlying connection is lost.
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -9,6 +9,7 @@ import { deserialize, serialize, RpcSession, type RpcSessionOptions, RpcTranspor
          newHttpBatchRpcSession} from "../src/index.js"
 import { swapByteOrder } from "../src/serialize.js"
 import { MAX_CLOSE_REASON_BYTES } from "../src/websocket.js"
+import { ErrorStubHook, PromiseStubHook, type StubHook } from "../src/core.js"
 import { Counter, TestTarget } from "./test-util.js";
 
 type CustomEncodingLevel = RpcTransportWithCustomEncoding["encodingLevel"];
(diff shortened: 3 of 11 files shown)


ting-hong-shieh:

kentonv:
Hi @ting-hong-shieh,

Thanks for trying to help.

PR #251, opened 2026-08-19
fix: align RpcPromise construction with result elision

ndisidore:
`new RpcPromise(promise)` now produces the same type as an RPC method returning the same thing.

Before, the two could disagree, which caused real problems:

- Wrapping a promised stub of a plain interface (the reconnect use case) inferred the wrong type. The awaited value was missing `dup()`, and constructor-made promises couldn't be mixed with ones returned by RPC methods.
- Inline object literals with methods, i.e. `Promise.resolve({ value, next() {…} })` didn't compile (less important probably but not an expensive rider)

The fix: the constructor's return type goes through the same stub elision as method results, and its parameter accepts either the payload or a stub of it. This also resolves @kentonv's [`| Stub<T>` question on #242](https://github.com/cloudflare/capnweb/pull/242#discussion_r3804971237).  Type tests pin that the constructor and method returns agree for every payload shape.

### DIFF
--- a/.changeset/rpc-promise-ctor-elision.md
+++ b/.changeset/rpc-promise-ctor-elision.md
@@ -1,0 +1,5 @@
+---
+"capnweb": patch
+---
+
+The `RpcPromise` constructor now applies the same stub elision as method result types: wrapping a `Promise<RpcStub<T>>` produces the same `RpcPromise<T>` a method declared to return that stub would, plain-interface stub payloads keep their stub type, and promises resolving to inline object literals with methods now infer correctly.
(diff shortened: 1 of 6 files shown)


__tests__/workerd.test.ts
```diff
@@ -205,7 +205,9 @@ describe("workerd compatibility", () => {
   })
 
   it("can wrap a userspace stub in a native stub", async () => {
-    let stub = new NativeRpcStub(new RpcStub(new JsCounter()));
+    // Cast: now that the `__RPC_TARGET_BRAND` no longer leaks onto stub surfaces, a userspace
+    // stub doesn't statically match workers-types' `Stubable` (runtime interop still works).
+    let stub = new NativeRpcStub(<any>new RpcStub(new JsCounter()));
```
ndisidore: ugh this is probably a non-starter
- capnweb stub → new NativeRpcStub(...): needs <any> now

but its because runtime suffers from the same poised branded stub issue

dimitropoulos (approved):

PR #253, opened 2026-08-19
fix: methods returning Promise<RpcStub<T>> no longer produce stub-of-stub result types

ndisidore:
Presently declaring an RPC method to return `Promise<RpcStub<T>>` gave you a broken type: an internal brand property made the type machinery treat the already-stubbed result as something that still needed wrapping, so awaiting yielded `RpcStub<RpcStub<T>>` (TS2322) and passing the result into another RPC call failed to compile (TS2345), even though the runtime behaved correctly. 

Now the type machinery recognizes existing stubs before re-wrapping, so a `Promise<RpcStub<T>>` return types identically to a `Promise<T>` return and awaits, pipelined calls, and .map() over stub arrays all compile.

Nothing changes at runtime; the one migration is rewriting any explicit `RpcPromise<RpcStub<T>>` annotation as `RpcPromise<T>`.

### DIFF
--- a/.changeset/result-stub-elision.md
+++ b/.changeset/result-stub-elision.md
@@ -1,0 +1,6 @@
+---
+"capnweb": minor
+"capnweb-validate": minor
+---
+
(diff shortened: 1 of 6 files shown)


dimitropoulos (approved):
type level tests, woop woop

PR #255, opened 2026-08-22
ci: deploy the docs (prod and preview)

dimitropoulos:
The docs site landed in #233 with a `wrangler.jsonc` and nothing that ran it. This points it at the domain and puts the deploy in CI.

**The site is already live at https://capnweb.com** -- I deployed it by hand from this branch's config to confirm the wiring before automating it. This PR is the config that produced it, plus the workflow that keeps it current.

## Setup required before this can deploy

Two things need repo-admin and Cloudflare dashboard access, so they are not in this PR:

1. **A Cloudflare API token** on the `capnweb` account, scoped to `Account / Workers Scripts / Edit`, `Zone / Workers Routes / Edit` on `capnweb.com`, and `Zone / Zone / Read` on `capnweb.com`.
2. **A `docs` GitHub environment** on this repo holding it as `CLOUDFLARE_API_TOKEN`. Separate from `release` on purpose -- docs deploys should not go through npm-publish approvals, and release approvals should not be spendable on a docs push.

Until both exist the `deploy` job fails at the last step. Nothing else is affected, and the currently deployed site stays up: Workers asset deploys are atomic, so a failed run leaves the previous version serving.

## Previews

Every pull request from a branch in this repo now gets its own copy of the site, via [Worker Previews](https://developers.cloudflare.com/workers/previews/) — `wrangler preview` against the same `capnweb-docs` Worker, so a preview is a branch of production rather than a second Worker to operate. **This PR's is at https://pr-255-capnweb-docs.capnweb-b14.workers.dev** (deployed by hand, since the secret does not exist yet).

`preview_urls` goes back to `true`. It is not the same switch as `workers_dev`, which my first commit had wrong: `workers_dev` is production's `workers.dev` copy, which we still do not want; `preview_urls` is what gives a preview a hostname at all. With it off, `wrangler preview` succeeds and returns a preview whose `urls` array is empty. Production stays off `workers.dev` either way — verified 404.

Two parts of `preview-docs.yml` are load-bearing:

- **The preview URL is derived before the build, not read back after it.** `site` is baked into canonicals, OG URLs, `robots.txt` and the sitemap at build time, so `DOCS_SITE_URL` has to be set *before* `astro build` or the preview publishes a sitemap claiming to be capnweb.com. The deploy step reads the URL back anyway and fails if it is not the one the site was built for.
- **`X-Robots-Tag: noindex` is appended to the built `_headers`,** not committed to `public/_headers`, which ships to production too. A header rather than a `robots.txt` `Disallow`, because disallowing the crawl would stop a crawler ever reading the noindex.

Fork PRs get no preview — no secrets, and `pull_request_target` would hand a contributor's build scripts a credential that can deploy the real site. Cleanup treats "preview not found" (`code: 10025`) as success, or every PR closing without one gets a red X.

### Dependency bump

`wrangler` 4.63.0 → 4.125.0, because `wrangler preview` is a private-beta command that does not exist in 4.63.0. That drags `@cloudflare/workers-types` v4 → v5 as a peer. Both are devDependencies and neither changes the published types; `test` passes in CI (node, workerd, all four browser projects, bun) and `test:types` is clean. `@cloudflare/vitest-pool-workers` stays at 0.12.10 on purpose — 0.22 requires vitest ^4, and a test-framework major is not a Workers dependency bump.

### Beta caveat

Worker Previews are in **private beta**, and two things in the current docs do not match the shipped CLI. This workflow works around both: `--json` has no `.preview_url` field (it is `.preview.urls[0]`, preceded by progress text on stdout), and `preview delete` takes `--name` as a flag — the positional is the entry point script, so `wrangler preview delete pr-1` tries to deploy a Worker called `pr-1`.


.github/workflows/preview-docs.yml
```diff
@@ -0,0 +1,172 @@
...
+  PREVIEW_NAME: ${{ github.event.pull_request.number }}
+  PREVIEW_URL: https://${{ github.event.pull_request.number }}.pr.capnweb.com
+
+jobs:
+  preview:
+    # A pull request from a fork gets a read-only token and no access to secrets, so it
+    # cannot deploy -- and `pull_request_target`, which would hand it both, would be
+    # running a contributor's build scripts with a credential that can deploy the real
+    # site. Fork pull requests are checked by the `build-docs` job in test.yml instead.
+    if: ${{ github.event.action != 'closed' && github.event.pull_request.head.repo.full_name == github.repository }}
```
ndisidore: nit: i always prefer stable repo is over `full_name` just to avoid string comparison
i.e. something like
```
github.event.pull_request.head.repo.id == github.event.pull_request.base.repo.id
```

ndisidore (approved):
Pushed a commit for some cleanup. lgtm!

PR #259, opened 2026-08-28
chore: enhance type to keep reference to its `T`

### LINKED ISSUE #258: Cannot "Go to Definition"
## Problem

When trying to navigate on a RPC stub in editor (Cursor), "Go to Definition" doesn't jump to its original server method.

It would be a nice enhancement if we can fix the type, so we can jump between files (definitions).

Example:

```ts
const api = newWebSocketRpcSession("...");
const result = await api.hello("world"); // 👈 I want to jump from `hello` to

class MyApiServer extends RpcTarget {
  hello(name) {                          // 👈 here
    return `Hello, ${name}!`;
  }
}
```

## How to reproduce

- Run the following from root to build capnweb and example repo:
  - npm run build
  - cd examples/worker-react/client
  - npm install
  - npm run build
- Open `examples/worker-react/client/src/main/runs.ts`
- Go to line 79 (`const user = api.authenticate('cookie-123'`), or any stub code)
- Right click on method name (in this case, `authenticate`) and select "Go to Definition"

Image

yusuke99:
Closes #258 

When creating a stub type from a server class `T`, it drop keys such as brand symbols. Currently it iterates with `Exclude<keyof T, …>`, which makes TypeScript then treats the result as a new object, so "Go to Definition" on `api.authenticate()` no longer jumps to `T.authenticate`. This PR still drops the same keys, but maps over `keyof T` first and then `Pick`.

Here is a minimal PoC difference between using `Exclude<keyof T, …>` vs using `Pick`:

https://www.typescriptlang.org/play/?#code/MYGwhgzhAEDKYFsAOICm0DeAoa0BmA9gQBQCUmOu0ATqgC4Cu1AdtHdQ6gNyUC+lAIzDUyFKjXpNW7Tj1z9+WOgE8k6AIIAeACoA+aAF4x0ANoBpaAEtWAUQAeoBgBNUmgNaplBPNG0AaaAByIWpA3QBdAC5fc3CeXh4lVXQAIR19IwAFS2A3TUpscXMrVg8vH20omLM4vj9Ke0cXd09vXwDg4TCsXUSXUGF0YAJmCDpoMGiteGQ0Xqx+8FpoYdHxgWi0mZRUeawwADpCElIeAHoz6AA9AH4sASOiMnPL26wsIA

https://github.com/user-attachments/assets/238e27a4-b262-4fcf-83ec-9c8f80d79fea

### DIFF
--- a/packages/capnweb-validate/src/internal/core.ts
+++ b/packages/capnweb-validate/src/internal/core.ts
@@ -219,16 +219,17 @@ type MapCallbackReturn<V> =
   InvalidNativePromiseInMapResult<V> extends never ? V : never;
 export type ValidatedStub<T> = MaybeCallableStub<T> &
   (T extends object
-    ? {
-        [K in Exclude<
+    ? Pick<
+        { [K in keyof T]: StubMethodOrProperty<T[K]> },
+        Exclude<
           keyof T,
           | symbol
           | "__RPC_TARGET_BRAND"
           | "__WORKER_ENTRYPOINT_BRAND"
           | "__DURABLE_OBJECT_BRAND"
           | keyof StubBase<never>
-        >]: StubMethodOrProperty<T[K]>;
-      } & {
+        >
+      > & {
         map<V>(callback: (value: MapCallbackValue<NonNullable<T>>) => MapCallbackReturn<V>): StubResult<
           Array<V>
         >;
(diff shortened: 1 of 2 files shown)


yusuke99:

ndisidore (commented):
Thanks for the DevEx wins and the included video! That's how you do a PR right there.
Looks good, just getting a few others on the team to take a look

ndisidore (approved):

PR #261, opened 2026-09-04
ci: tolerate missing preview URL output

yomna-shousha:
## Summary
- tolerate Wrangler preview output that omits preview/deployment URL arrays
- derive fallback preview and deployment URLs from preview_slug/deployment_id while keeping custom preview names working

## Verification
- reproduced the existing jq failure with successful preview output missing URL arrays: exit=4
- verified the updated jq filter against failed-log output shape
- verified it still handles previous output with preview_urls/deployment_urls
- verified custom preview_slug fallback output
- ran git diff --check


yomna-shousha:
Closing this because deriving preview URLs in the workflow would mask the real problem: Wrangler is not returning preview/deployment URLs, and the expected preview hostnames do not resolve.

PR #262, opened 2026-09-04
ci: enable docs preview domain

yomna-shousha:
## Summary
- keep `capnweb.com` as the production custom domain
- add `pr.capnweb.com` as a preview-enabled custom domain route
- keep bare `pr.capnweb.com` disabled while allowing preview URLs like `https://<preview-name>.pr.capnweb.com`

## Verification
- tested the same config shape on a disposable Worker using `testpr.yomnashousha.com` and `pr.testpr.yomnashousha.com`
- `npx wrangler deploy` reported `pr.testpr.yomnashousha.com (custom domain) [disabled, previews: enabled]`
- `npx wrangler preview --name smoke2` returned `preview_urls: ["https://smoke2.pr.testpr.yomnashousha.com"]`

## Follow-up
This must be applied by `npx wrangler deploy -c packages/docs/wrangler.jsonc`; `wrangler preview` does not create/update custom domain routes. The existing `Deploy Docs` workflow runs that command on `main`, so this should be provisioned after merge when the docs deploy runs.

### DIFF
--- a/packages/docs/wrangler.jsonc
+++ b/packages/docs/wrangler.jsonc
@@ -17,6 +17,7 @@
 		{
 			"pattern": "pr.capnweb.com",
 			"custom_domain": true,
+			"enabled": false,
 			"previews_enabled": true
 		}
 	],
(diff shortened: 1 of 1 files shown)


dimitropoulos:
thanks - added in https://github.com/cloudflare/capnweb/pull/257/changes/a0219851dec34e3d4667d8e63b4bed973cc8676c

PR #264, opened 2026-09-08
Fix deserialization of persisted byte maps

kaiicheng:
﻿Fixes a deserialization crash when a raw `Uint8Array` crosses a JSON persistence or transport boundary and is restored as a numeric-key object.

The decoder now accepts only ordinary or null-prototype objects whose keys are exactly contiguous `"0"` through `"n"` and whose values are integer bytes in `0..255`. It continues to reject gaps, extra keys, fractions, strings, nulls, and out-of-range values. Existing Uint8Array, base64, and byte-container marker handling are unchanged.

This is a compatibility fix for consumers using `jsonCompatibleWithBytes`; it does not change the preferred transport behavior of preserving byte arrays or encoding them as base64.

Related: openai/codex#42760

Validation:
- `npx vitest run --project node __tests__/index.test.ts __tests__/limits.test.ts` (222 passed)
- `npm exec --yes --package=node@24 -- npm run build`

### DIFF
--- a/.changeset/blue-bytes-restore.md
+++ b/.changeset/blue-bytes-restore.md
@@ -1,0 +1,5 @@
+---
+"capnweb": patch
+---
+
+Restore valid numeric-key byte maps produced by JSON persistence.
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -4211,3 +4211,66 @@ describe("deserialization and transport correctness", () => {
     expect(sentReason).toBe("a".repeat(MAX_CLOSE_REASON_BYTES - 1));
   });
 });
+
+describe("restored numeric byte maps", () => {
+  it("restores contiguous bytes including both boundaries", () => {
+    expect(deserialize('["bytes",{"2":255,"0":0,"1":123}]'))
+        .toEqual(new Uint8Array([0, 123, 255]));
+  });
+
+  it("restores empty bytes", () => {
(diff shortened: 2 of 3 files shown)


kaiicheng:

kentonv (changes requested):
It seems to me that you shouldn't be using `jsonCompatibleWithBytes` encoding level if your transport is not actually able to round-trip a `Uint8Array`. You should be using plain `jsonCompatible`. This is literally the only difference between them: `jsonCompatible` encodes bytes as base64 strings, which is really what you want in this case, as it will be much more efficient.

Can your use case switch to `jsonCompatible` mode instead? If not, can you elaborate on how you got into this situation, where you are using `jsonCompatibleWithBytes` but your transport doesn't actually support bytes?

kaiicheng:
Thanks for the clarification. I agree that reconstructing numeric-key objects in the decoder is not the right fix, so I'm withdrawing this PR.

The affected Codex extension bundle selects jsonCompatibleWithBytes on both sides of its VS Code RPC connection. However, further inspection shows that VS Code supports typed-array serialization, so postMessage alone does not establish where the type is lost. I have not yet isolated the failing boundary or validated an encoding change through a real restart/reconnect.

I'll continue the downstream investigation in openai/codex#42760. Thank you for pointing me toward the correct transport contract.

Issue #266, opened 2026-09-09
A single undecodable inbound frame aborts the whole session: `readLoop` has no per-frame error boundary


Kanishkrawatt:
### Summary

`readLoop` decodes every inbound frame under a single session-wide error handler, so any one frame that fails to decode takes down the whole session: every outstanding import is rejected and every export disposed. A per-frame boundary would make most of these failures cost one call instead of the session.

Observed on `0.12.0`.

### Where

`readLoop` (`dist/index.js:2236`) is a bare `while` loop around a `switch` on the frame type, and the only handler is at the call site:

```js
this.readLoop().catch((err) => this.abort(err))   // :1929
```

Anything any arm throws reaches `abort()`. Concretely:

- A bad export id in a `push` argument: `no such entry on exports table` (`:1586`, `:1618`, `:1628`).
- A duplicate or underflowing `release`: `no such export ID` (`:1969`, `:1983`) and `refcount would go negative` (`:1970`).
- An over-limit value: `maxDepth` (`:1437`) and `maxBigIntDigits` (`:1451`).

### Repro

Any peer bug that emits an undecodable frame will do. The one we hit is #265, a distinct defect in `getImport()`; its transcript shows the shape, where a single `["pipeline",1]` naming a released id ends the session:

```
client -> ["push",["pipeline",0,["use"],[["pipeline",1]]]]
client -> ["pull",3]
server -> ["abort",["error","Error","no such entry on exports table: 1"]]
SESSION DEAD: no such entry on exports table: 1
```

A limit violation behaves the same way, which is easier to reproduce standalone:

```js
const srv = new RpcSession(s.t, new Api(), { limits: { maxDepth: 8 } })
const cli = new RpcSession(c.t)
const api = cli.getRemoteMain()

await api.getThing()               // fine

let deep = {}, cur = deep
for (let i = 0; i < 40; i++) { cur.a = {}; cur = cur.a }
await api.use(deep)                // TypeError: Deserialization exceeded maximum allowed message depth of 8.

await api.getThing()               // SESSION DEAD, same error
```

We understand the limits case is deliberate: `RpcSessionOptions.limits` documents it as "A message that exceeds a limit is rejected, aborting the session" and describes the limits as guarding "against resource-exhaustion attacks from untrusted peers" (`dist/index.d.ts:279-285`). It is included here only because it demonstrates the single handler, not because we think the documented behaviour is wrong. See the last section for why that makes this an opt-in question.

### Suggested fix

[...]

PR #267, opened 2026-09-09
Don't cite a released import id after the reference has settled

### LINKED ISSUE #265: `getImport()` returns the id of an already-settled import, sending a released id to the peer
### Summary

`getImport()` returns an import's id without checking whether that import has already settled and been released. Passing a settled `RpcPromise` as an argument therefore puts a dead id on the wire, and the peer aborts the entire session.

Reproduced on `0.12.0` (also on `0.10.0`) with both peers on identical code, so version skew is not required.

### Repro

```js
import { RpcSession, RpcTarget } from 'capnweb'

// in-memory transport pair, logging every frame
const mk = (label) => { const q = []; let w = null
  return { push: (m) => { if (w) { const r = w; w = null; r(m) } else q.push(m) },
    t: { send(m) { console.log(`${label} ->`, m); peer[label === 'client' ? 'server' : 'client'].push(m) },
         receive() { return q.length ? Promise.resolve(q.shift()) : new Promise(r => { w = r }) },
         abort(e) { console.log(`${label} transport abort:`, String(e)) } } }
}
const c = mk('client'), s = mk('server'); const peer = { client: c, server: s }

class Api extends RpcTarget {
  getThing() { return { n: 1 } }
  use(x)     { return { echoed: x } }
}
const server = new RpcSession(s.t, new Api())
const client = new RpcSession(c.t)
const api = client.getRemoteMain()

const p = api.getThing()
const v = await p

await api.use(v)   // fine
await api.use(p)   // passes the same promise object, not its value
await api.getThing()   // session is gone
```

### Actual output

```
--- await, then pass the VALUE ---
client -> ["push",["pipeline",0,["getThing"],[]]]
client -> ["pull",1]
server -> ["resolve",1,{"n":1}]
client -> ["release",1,1]
v = { n: 1 }
client -> ["push",["pipeline",0,["use"],[{"n":1}]]]
client -> ["pull",2]
server -> ["resolve",2,{"echoed":{"n":1}}]
client -> ["release",2,1]
use(v) = { echoed: { n: 1 } }

--- await, then pass the same PROMISE object ---
client -> ["push",["pipeline",0,["use"],[["pipeline",1]]]]
client -> ["pull",3]
server -> ["abort",["error","Error","no such entry on exports table: 1"]]
server transport abort: Error: no such entry on exports table: 1
client transport abort: Error: no such entry on exports table: 1
use(p) THREW: no such entry on exports table: 1

--- is the session still usable? ---
SESSION DEAD: no such entry on exports table: 1
```

The `use(p)` frame is the bug: `["pipeline",1]` names import `1`, which the client itself released two frames earlier with `["release",1,1]`.

### Expected

Passing a settled `RpcPromise` as an argument should either resolve locally from the value already held, or fail that one call. It should not end the session. Note that passing the awaited *value* (`use(v)`) has always worked; passing the *promise object* (`use(p)`) is the trigger, and the two are easy to confuse in application code.

### Cause

This looks like a missed case in an established pattern rather than a missing pattern.

`ImportTableEntry.resolve()` stores the value in `this.resolution` and immediately calls `this.sendRelease()` (`dist/index.js:1750`, `:1756`), which sends the release frame and zeroes `remoteRefcount` (`:1806`). Both sides then correctly agree the id is dead. The id itself has to survive settlement, because `sendRelease()` reads `this.importId` to name what it releases. What should not survive is its *use*.

Five accessors already know this and operate on the settled value instead of the wire:

| accessor | line (0.12.0 `dist/index.js`) |
|---|---|
| `call` | `:1839` |
| `stream` | `:1844` |
| `map` | `:1851` |
| `get` | `:1856` |
| `pull` | `:1865` |

`getImport()` (`:2036`) does not. It asks only whether the hook belongs to this session:

```js
getImport(hook) {
  if (hook instanceof RpcImportHook && hook.entry && hook.entry.session === this) return hook.entry.importId
  else return
}
```

The resulting frame is fatal on the peer because `evaluateImpl` throws `no such entry on exports table` (`:1586`, `:1618`, `:1628`) inside `readLoop`, whose only error handler is session-wide: `this.readLoop().catch((err) => this.abort(err))` (`:1929`). That second half is #266, filed separately since it is a distinct decision; this issue is about not emitting the bad frame in the first place.

### Suggested fix

Extend the same `entry.resolution` gate to `getImport()`, so a settled entry withholds its id and the caller's existing `undefined` path takes over (the serializer calls it at `:1329` and already branches on `importId !== void 0` at `:1330`, which is the site that emits the `["pipeline", importId, ...]` seen above).

`awaitResolution()` (`:1772`) has the identical shape, calling `sendPull(this.importId)` without testing `this.resolution`. It appears unreachable in that state: its only two callers are `RpcImportHook.pull()`, which returns on `entry.resolution` one line earlier at `:1865`, and the `sendStream` path at `:2162`, whose entry is constructed with `pulling = true` so the constructor seeds `activePull` and the `sendPull` branch never runs. Guarding it costs one line and cannot be covered by a test for that reason. Worth flagging either way, since the shape remains.

### Impact

Six users across macOS and Windows since 2026-07-31, in an app where one session multiplexes every backend service, so a single occurrence costs the user every service at once and forces a reload.

Happy to open a PR if the approach looks right.

Kanishkrawatt:
Fixes #265.

### The problem

`ImportTableEntry.resolve()` stores the resolution and immediately calls `sendRelease()`, so the import is dead on both sides. The entry keeps `importId` — it has to, because the release accounting names the id through it — but `getImport()` still returned that id unconditionally.

So passing a settled `RpcPromise` back as an argument re-serialized a released id. The peer cannot find it and throws inside `readLoop`. Since `readLoop` is wrapped in a single session-wide `.catch(err => this.abort(err))`, a call-level fault tears down the **whole session**, not just the offending call.

We hit this in production on a connection that multiplexes every service in an Electron desktop app: one stale reference took the entire connection down and the user got a "connection lost" screen.

### The fix

`getImport()` declines a settled entry, so the caller exports a fresh stub instead. This is the same test `dispose()`, `abort()` and `onBroken()` already apply — they all branch on `resolution`; `getImport()` was the one that did not.

The guard gates the *use* of `importId` rather than clearing it, so release accounting is unchanged. A test covers that specifically, because a guard that cleared the id would still emit a release frame — just one naming `null` — and silently leak the peer's exports.

`awaitResolution()` gets the same guard. It is **not reachable** with a settled entry today: `RpcImportHook.pull()` returns on `entry.resolution` one line earlier, and `sendStream` seeds `activePull` via `pulling = true`, so the `sendPull` branch never runs. It is defensive only — included so a future caller cannot regress into a `pull` naming a released id. Happy to drop it if you would rather not carry an unreachable branch.

### Tests

Two tests in `__tests__/index.test.ts`. The first is load-bearing — verified to **fail on unpatched `main`** with `no such entry on exports table`, and to pass with the fix. The second guards against the clear-the-id mis-fix described above.

Full suite on this branch:

| suite | result |
|---|---|
| `vitest run --project=node` | 407 / 407 pass |
| `vitest run --project=workerd` | 210 / 210 pass |
| `npm run test:types` | pass |

I did not run the browser projects or `test:bun` locally (Playwright download blocked behind a corporate proxy); neither covers this path, but CI will.

### Scope note

There is a second, separate defect where the settled resolution is a **non-promise stub** — the library then cannot pull it and the call fails with `Tried to resolve a non-promise stub`. That is unchanged by this PR and is out of scope here; this change only stops that case from being *session*-fatal rather than call-level. Reported separately in #266, which is about `readLoop`'s missing per-frame error boundary.

Present in 0.10.0 and still in 0.12.0, so upgrading is not a workaround — we are currently carrying this as a local patch against the published bundle, which is what prompted the report.

### DIFF
--- a/src/rpc.ts
+++ b/src/rpc.ts
@@ -247,6 +247,11 @@ class ImportTableEntry {
   }
 
   async awaitResolution(): Promise<RpcPayload> {
+    // If the entry has already settled, the import has been released (resolve() calls
+    // sendRelease()), so there is nothing left on the wire to pull. Read the stored
+    // resolution instead of sending a "pull" naming a released id.
+    if (this.resolution) return this.resolution.pull();
+
     if (!this.activePull) {
       this.session.sendPull(this.importId);
       this.activePull = Promise.withResolvers<void>();
@@ -662,7 +667,11 @@ class RpcSessionImpl implements Importer, Exporter {
   }
 
   getImport(hook: StubHook): ImportId | undefined {
-    if (hook instanceof RpcImportHook && hook.entry && hook.entry.session === this) {
+    // A settled entry has already released its import (resolve() calls sendRelease()), so its
+    // importId no longer names anything on the peer. Fall through to exporting the resolution,
+    // the way dispose(), abort() and onBroken() all branch on `resolution`.
+    if (hook instanceof RpcImportHook && hook.entry && hook.entry.session === this &&
+        !hook.entry.resolution) {
       return hook.entry.importId;
     } else {
       return undefined;
--- a/.changeset/settled-import-guard.md
+++ b/.changeset/settled-import-guard.md
@@ -1,0 +1,14 @@
+---
+"capnweb": patch
+---
+
+Don't cite a released import id after the reference has settled.
+
+`ImportTableEntry.resolve()` stores the resolution and immediately calls `sendRelease()`, so the
+import is dead on both sides — but the entry keeps `importId`, since the release accounting names
+the id through it. `getImport()` still returned that id, so passing a settled `RpcPromise` back as
+an argument re-serialized a released id. The peer could not find it and threw inside `readLoop`,
+which is wrapped in a single session-wide `.catch(err => this.abort(err))` — so a call-level fault
+destroyed the whole session. `getImport()` now declines a settled entry and the caller exports a
+fresh stub instead, matching `dispose()`, `abort()` and `onBroken()`, which already branch on
+`resolution`.
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -4227,3 +4227,44 @@ describe("deserialization and transport correctness", () => {
     expect(sentReason).toBe("a".repeat(MAX_CLOSE_REASON_BYTES - 1));
   });
 });
+
+describe("settled import references", () => {
+  // Regression: ImportTableEntry.resolve() stores `resolution` and immediately calls
+  // sendRelease(), so the import id is dead on both sides -- but the entry keeps `importId`,
+  // because the release accounting names the id through it. getImport() then still handed out
+  // that dead id, so passing the settled promise object into a later message re-serialized a
+  // released id. The peer could not find it and threw inside readLoop, and because readLoop is
+  // wrapped in a single session-wide `.catch(err => this.abort(err))`, a call-level fault
+  // destroyed the entire session.
+  //
+  // dispose(), abort() and onBroken() all already branch on `resolution`; getImport() did not.
+  it("does not cite a released import id after the reference has settled", async () => {
+    let harness = new TestHarness(new TestTarget());
+
+    // Awaiting settles the entry: the resolution is stored and the import is released.
+    let promise = harness.stub.returnNumber(7);
+    await promise;
+
+    // Passing the same *promise object* (not its awaited value) as an argument is what
+    // re-serialized the now-dead id.
+    expect(await harness.stub.square(promise)).toBe(49);
+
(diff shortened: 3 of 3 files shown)


Kanishkrawatt:

ndisidore:
@ask-bonk review this

src/rpc.ts
```diff
@@ -247,6 +247,11 @@ class ImportTableEntry {
   }
 
   async awaitResolution(): Promise<RpcPayload> {
+    // If the entry has already settled, the import has been released (resolve() calls
```
ndisidore: This hunk can be safely removed as I have confirmed there are no code paths able to actually hit this

